PR #50, opened 2017-01-19
Open mp features - reduced annotations

SaeedDarabi:


SaeedDarabi:
OpenMP section example with reduced annotations.

PR #124, opened 2018-09-17
Fix to make the Windows build find the viper files.

HenkMulder:


sjcjoosten:
Error is in the travis system are not due to this commit, so I'm accepting

PR #136, opened 2018-11-20
Add rewrite rule to drop useless quantifier

HenkMulder:


sjcjoosten:
This rewrite rule is valid for ``\forall``, but not for the separating forall:

``
\forall* int i; ( i \memberof [ 0 .. n )); Perm(x,i/n)
``

is equivalent to

``
n > 0 ==> Perm(x,1)
``

I propose not to merge and delete this branch.

Issue #111, opened 2018-05-24
The 'ant dist' target still uses Scala 2.11.7 


wytseoortwijn:
https://github.com/utwente-fmt/vercors/blob/a2f042ced7df2fd1f6a29647796c952c25466164/build.xml#L56

Upgrade to Scala 2.11.8.

wytseoortwijn:
Fixed in #141 (upgraded to Scala 2.12.7)

PR #141, opened 2018-12-17
fixes #137: update build.xml

### LINKED ISSUE #137: JAR assembly does not run/build correctly
It seems that at least it cannot find the environment variables set in `unix/bin/vct`

niomaster:
Please note the version bump of scala-library


niomaster:
Probably also fixes #111

PR #142, opened 2018-12-17
Viper update

wytseoortwijn:


wytseoortwijn:
Travis seems to remain unstable, also after the Viper update. I propose we ignore the failing examples for now (on my machine they pass) and resolve the instability of Travis in the context of #143.

Issue #126, opened 2018-09-26
Java.lang exception


DanTUe:
The following program gives a java exception stack trace:

```
class Problem {
	int[] array;
}
```

sjcjoosten:
Thank you so much for finding this! The problem is with the ``int[] array;`` on the second line, but VerCors should of course tell you that the problem is here with a better error message!

wytseoortwijn:
Perhaps this is clear already, but VerCors currently requires all fields of non-primitive type to be assigned some value in a constructor. For example, the following adaption would verify:

```java
class Problem {
	int[] array;
	
	Problem(int[] a) {
		array = a;
	}
}
```

The assignment `array = null;` would also be fine. So one could either improve the error messages, or simply allow such fields to be null (e.g. as a default), thereby avoiding such stack traces.

sjcjoosten:
The option of using null by default is the one we prefer

sjcjoosten:
Is this issue fixed as well, now that #127 is fixed?

HenkMulder:
This issue has nothing to do with the RewriteArrayRef class. The problem is in initialization of an int[] field by an implicit constructor.

VerCors encodes an int[] field as VCTOption<Seq<Ref>>. In the implicit constructor that is generated by VerCors this field is being set to null and the value of this field is being compared with null. Which is wrong, since it should be set to, and compared to VCTNone().

niomaster:
Fixed in #156

PR #147, opened 2018-12-19
Array initialization and null assignment.

HenkMulder:
Fix for issue #126 and null-assignment to arrays.


niomaster:
This is fixed by the new array format and related code, so let's discard these changes and use that instead.

PR #152, opened 2019-01-21
Add for loop to PVL

### LINKED ISSUE #107: For-loops in PVL
For-loops should be implemented and supported in PVL. Currently only while-loops are implemented. The implementation of for-loops should be a straightforward translation to while-loops.

niomaster:
fixes #107 

I do have a question: in the visit method for OperatorExpression in the Standardize pass a check is made whether the parent node is a BlockStatement. This seems to guard that several types of expressions are not standardized before they are flattened, so side effects only occur on the top level. However, not rewriting e.g. postfix increments breaks the final translation to silver regardless, so the checks seems unnecessary, or am I missing something?


niomaster:
I see a bunch of tests fail, so I'll check that out first and close the pull request for now.

PR #155, opened 2019-02-18
Change viper-api to sbt multi-project

HenkMulder:
* Move viper folder to vercors/viper.
* Move all viper-api related source files to the new vercors/viper root
project (of the vercors/viper sbt multi-project).
* Change vercors project to an sbt multi-project, that depends on
viper-api.
* Update SilverBackend to use normal constructors instead of jars to get
the verifiers.
* Update installation instruction and build- and run scripts.

In the new project structure the carbon, silicon and silver projects are
now one-to-one copies of their respective viper repositories. See
vercors/viper/README.md.

Has a new version of Viper. Also contains the changes from #133 and the
viper-api related changes from #151.


sjcjoosten:
Pieter, you have a unix platform right? Can you try this out there? Does this branch also cause unit tests to run?

niomaster:
The build fails on viper/carbon for me, so I can't really test whether unit tests are executed. However, I found the unit test condition to be in the root build.xml file, see here: https://github.com/utwente-fmt/vercors/commit/693bba8127012520a7ae75cf7dcd3c4e03dbdaea#diff-2cccd7bf48b7a9cc113ff564acd802a8L11

HenkMulder:
Make sure that you have created the symbolic links linking the carbon and silicon modules to silver. Without them the build will fail indeed. 
See the README file in vercors/viper.

niomaster:
Yes, that does work. However, if you first try to build the project without creating the symlinks, the build script creates two directories in the same location where the symlink should be. In that case the symlink is created in the wrong place (within the wrongly created empty directory), so maybe it's worth it to just create a small script that checks whether those directories were incorrectly generated and create the platform-dependent symlinks?

Also, I noticed that assembly is broken, as the required resources seem to be in a different place now :)

HenkMulder:
Let me update the travis_build.sh script. Then Unix users can also use that to remove erroneous directories, create symbolic links and build VerCors.
Could you elaborate on the assembly part? Something in an ant task?

niomaster:
yes, the assembly task of the top-level build.xml

HenkMulder:
'ant dist' should now also work on both Windows and Unix.

Issue #153, opened 2019-01-21
Java arrays should be rewritten to Option[Array[T]]


niomaster:
Java arrays should make use of the upcoming new rules for the Array type in Col.

PR #156, opened 2019-02-20
Rewrite Java & PVL arrays to new Array type

niomaster:
Fixes Issue #153. Work is not complete yet as some expression cases are missing, and Java 8 is not fixed yet.

Should be merged in conjunction with fixes to rewrite_arrays and recognize_multidim.


niomaster:
Java 8 is now supported, as are all operators, except casts. As far as I'm concerned the Java translation part is now done, so this is ready as soon as the work of @HenkMulder is complete.

HenkMulder:
The changes break 2 unit tests, which causes a build on Windows to fail.

Further most array examples don't get to the rewrite_arrays step. I tried
* examples/manual/array.pvl
* examples/arrays/JavaArrayExamples.java
* examples/arrays/basic-examples.c

niomaster:
Hi, thank you for your testing!

I would expect the pvl and c examples to fail, as I have had to restrict the type checker in order to be able to rewrite null values using type information. In particular, the examples fail as null is not allowed as a value of arrays, so where appropriate the pvl/c parsers should rewrite the arrays to option[array[]] etc.

The Java example is actually quite interesting, it seems that array.length is rewritten to Size(array), although I was under the impression that was the job of the rewrite_arrays stage. I'll investigate that further.

Finally, thank you for bringing the unit tests to my attention. We made a decision to not support Object to be a supertype of non-class types, so that's why a unit test failed. It's fixed in the latest commit.

niomaster:
I cannot find further failing Java examples due to the array change, so let me know if you still find anything.

HenkMulder:
Hi. In #158 are the changes to use the new array encoding for Viper.
The RewriteArrayRef class in this version modifies the PrimitiveType array, and does all other rewriting based on the new type. Thus if all array related types and operations are set correctly you could remove the visit(PrimitiveType t) method from RewriteArrayRef.java, and then everything should work.

PR #158, opened 2019-03-04
Vctarray

HenkMulder:
Use new array encoding in Viper. Speeds up verification with \array and \matrix annotations. Multi-dimensional arrays (matrices) are no longer flattened to a one-dimensional Sequence. This avoids having to do a (non-linear) calculation to address a element of a multi-dimensional array.
This also fixes some some failing examples with arrays.


vercors/src/main/java/vct/col/rewrite/RewriteArrayRef.java
```diff
@@ -3,269 +3,391 @@
...
+      }
+      if (array != null && rewrite(array.getType()).isPrimitive(PrimitiveSort.Option)) {
+        result = create.expression(e.operator(), create.reserved_name(ASTReserved.OptionNone), rewrite(array));
+      } else {
+        super.visit(e);
+      }
+      break;
+    }
+    case Length:
+      if (e.first().getType().isPrimitive(PrimitiveSort.Array)) {
```
niomaster: This should probably be the other way around, no? If it is an option you should generate a get, when it is an array you shoud not.

HenkMulder:
Note that we check the old type. In this version all array types are rewritten to option types, so for this version it is correct. For your work this would indeed cause a failure. I think that in combination with your work you could remove the entire case for Length, since then the type would not change anyway.

PR #163, opened 2019-03-19
Matcher for rewrite passes

### LINKED ISSUE #160: More ergnomic rewrite-pass unit tests
We should investigate whether it's possible to make writing unit tests for rewrite passes easier, as it often should contain checks of the form: is this AST rewritten to this other AST.

niomaster:
fixes #160: adds RewriteSpec, which can match trees for rewrite passes. The example tests works within #156, but we can cherry-pick `vct.rewrite.RewriteSpec` if we want to merge this earlier.


niomaster:
In the Java 7 and PVL parsers the line is unreachable based on the current grammar. In the Java 8 case a different syntax is actually not covered. In any case, the error should probably be a little bit more descriptive.

PR #186, opened 2019-04-12
fixes #61: Don't allow flattening some terminals: constants, names

### LINKED ISSUE #61: Returning `null` in PVL
One can currently not return `null` in a method body in PVL. This results in an exception with a message that looks like: `cannot declare variable __flatten_? of <<null>> type.` (with `?` some integer).

Minimal example program (in PVL):

```java
class List {
  List nothing() {
    return null;
  }
}
```

niomaster:


niomaster:
That's an interesting thought, I think forbidding to flatten any terminal would give rise to that property, but I'm fairly certain these two do not cover all terminal expression nodes.

sjcjoosten:
Then let's not add failing (or succeeding but misleading) tests just yet, and see if the need to do this ever arises.

PR #192, opened 2019-04-25
fixes #78: use the logging framework everywhere, more restrictive in log level

### LINKED ISSUE #78: Too much output
Vercors is showing too much output. It would be nice if it would, at least by default, give only the errors in your program.

niomaster:
Adds configurable logging levels to the hre logging framework, plus command line options to manage them.


niomaster:
It seems that I've made the travis test so silent that the build is killed by travis...

niomaster:
I'm going to need some time to diagnose the travis issue unfortunately, as I cannot easily reproduce the failure in isolation (only when running the full test suite)

vercors/build.sbt
```diff
@@ -35,7 +34,7 @@ lazy val vercors = (project in file("."))
 
     assembly / mainClass := Some("vct.main.Main"),    // Define JAR's entry point
     assemblyMergeStrategy in assembly := {
-      case "logback.xml" => MergeStrategy.first
+      case "logback-test.xml" => MergeStrategy.first
```
sjcjoosten: I don't understand the "logback-test.xml" case, why is it there / what is it for? Can we put this in a comment?

niomaster: For documentation's sake, the file `logback-test.xml` does exist. It is there as a hack to override the logging configuration in `viper`, so it goes through `hre`. The merge strategy shouldn't have said `logback-test.xml`, as it is there to consolidate the logging configuration of viper, which is just called `logback.xml`.

PR #201, opened 2019-05-06
fixes #194: forbid calling non-pure methods in pure methods

### LINKED ISSUE #194: Local variable access in pure method gives "current thread id not defined" error
I have the following program:
```
class MyClass {
  int[] P;

  requires Perm(P, 1);
  requires P != null;
  requires (\forall* int i; 0 <= i && i < P.length; Perm(P[i], 1));
  requires (\forall int i; 0 <= i && i < P.length; 0 <= P[i] && P[i] < P.length);
  pure int getPure(int v) = get(P[v]);

  requires Perm(P, 1);
  requires P != null;
  requires (\forall* int i; 0 <= i && i < P.length; Perm(P[i], 1));
  requires (\forall int i; 0 <= i && i < P.length; 0 <= P[i] && P[i] < P.length);
  int get(int v) {
    return P[v];
  }
}

/*
class MyClass2 {
  int[] P;

  pure int getPure(int v) = get(P[v]);

  int get(int v) {
    return P[v];
  }
}
*/
```

Vercors gives the following error:

```
=== global-pure-issue.pvl          ===
   6   requires (\forall* int i; 0 <= i && i < P.length; Perm(P[i], 1));
   7   requires (\forall int i; 0 <= i && i < P.length; 0 <= P[i] && P[i] < P.length);
                                [---------
   8   pure int getPure(int v) = get(P[v]);
                                 ---------]
   9 
  10   requires Perm(P, 1);
  Local name current_thread_id is undefined
```

If the bottom class is uncommented it complains that ```Local name globals is undefined```. I am probably not using the language constructs correctly (am I?) but the error was somewhat puzzling to me nonetheless, since I am not using either variable name in my programs.

niomaster:


niomaster:
@wytseoortwijn will take a look at the potentially broken examples.

sjcjoosten:
Quite a number of tests fail after forbidding this:

```
  ConcurrentCounting/silicon: verdict is Error instead of Pass
  GeneralisedConcurrentCounting/silicon: verdict is Error instead of Pass
  ListAppendASyncDef/silicon: verdict is Error instead of Pass
  ListAppendASyncDefInline/silicon: verdict is Error instead of Pass
  TreeWandSilverE1/silicon: verdict is Error instead of Fail
  TestScale/silicon: verdict is Error instead of Fail
  SimpleLockingProtocol/silicon: verdict is Error instead of Pass
  FunctionsPVL/silicon: verdict is Error instead of Pass
  TreeWandSilverE2/silicon: verdict is Error instead of Fail
  NoSendAfterRead/silicon: verdict is Error instead of Pass
  HistoryApplication/silicon: verdict is Error instead of Pass
  ListsPVL/silicon: verdict is Error instead of Pass
  HistoryLoop/silicon: verdict is Error instead of Pass
  UnequalCounting/silicon: verdict is Error instead of Pass
  ListAppend/silicon: verdict is Error instead of Pass
  ParallelGCD/silicon: verdict is Error instead of Pass
  TreeRecursiveSilver/silicon: verdict is Error instead of Pass
  WandDemoSilver/silicon: verdict is Error instead of Pass
```

Maybe we should discuss what to do with these and how to fix these on a VerCors meeting.

niomaster:
I don't think we're realistically going to solve this soon, so should we close the PR for now?

sjcjoosten:
I'm assigning Raul to make a decision on that

raulmonti:
OK I'll come back to this next week and see what to do. We may be able to talk about it if we do a VerCors meeting even though Marieke is not here.

wytseoortwijn:
This PR current seems to be approved, but be aware that merging it breaks support for histories and futures.

niomaster:
I've changed the check as follows:

- If we are in a pure method and we encounter a method invokation, we check that either:
     - The invoked method is pure
     - The "invoked method" is a predicate
- If neither is true, we make an exception when we are in a method that defines a process, as for now that is very broken in vercors: some process definitions are marked pure because they are on the defintion level, whereas this property is stripped in some places because on the process algebra level they are not pure.

raulmonti:
>     * as for now that is very broken in vercors:

Should we make a new issue for fixing this? so we do not forget about the problem after merging this request.

niomaster:
Yes, although I think that falls under the larger issues of "we want a DSL for models at some point"

PR #203, opened 2019-05-08
added partial solutions for verifythis2019

wytseoortwijn:


wytseoortwijn:
Sure, I'll do this. Let's not put them in the Travis suite for now.

PR #211, opened 2019-05-14
Added Challenge 2b of VerifyThis'19

wytseoortwijn:


niomaster:
This broke the travis build because 2b is unmarked

wytseoortwijn:
I'm afraid so, but PR #212 fixes it again.

PR #214, opened 2019-05-15
fixes #210: only restore cell type when dereferencing the cell item

### LINKED ISSUE #210: Cannot access member objects within array
In the following code `bar` is verified, but `foo` fails with the "val is not a pseudo field" error:

```
class Int {
  int val;
}

class MyClass {
  Int[] xs;
  seq<Int> ys;

  requires xs.length == 1;
  int bar() {
    return ys[0].val;
  }

  requires Perm(xs, 1);
  requires xs != null;
  requires xs.length == 1;
  requires Perm(xs[0], 1/2);
  int foo() {
    return xs[1].val;
  }
}
```

Output:
```
Plain bar
Plain foo
Parsed 1 file(s) in: 229ms
Comparing Integer with 0 as Integer
At file pseudo-field-array.pvl from line 19 column 12 until line 19 column 20: val is not a pseudo field of (Cell<Int>).
Triggered by file pseudo-field-array.pvl from line 19 column 12 until line 19 column 20:
The final verdict is Error
entire run took 301 ms
```

This was fixed for sequences in #190 but the fix does not apply to arrays it seems, sadly. I tried looking at the code to see if I could generalize it, but I don't know where to start.

niomaster:


sjcjoosten:
Travis failure seems unrelated and may be fixed after the merge.. should we just try?

niomaster:
All tests passed except the unmarked one that Wytse fixed, so I think we can just merge this.

PR #221, opened 2019-05-29
Setup Sonar scan

Sophietje:


Sophietje:
It works! 
If you want to have a look, go to the VerCors SonarCloud page (https://sonarcloud.io/dashboard?id=utwente-fmt_vercors) and select the PR 221 branch at the top (where it says 'master'). You should then end up here: https://sonarcloud.io/dashboard?id=utwente-fmt_vercors&pullRequest=221
Note that at this point master has not yet been analyzed because this PR has not been merged. Therefore, it is correct that there are no results for this yet.

PR #229, opened 2019-06-14
Generate triggers for universal quantifiers

HenkMulder:
Generate triggers for universal quantifiers.

Rewrite complex subscripts (subscripts that are not allowed in trigger expressions) in quantifiers and (if possible) generate triggers for universal quantifiers (forall and forall*).

Use the Integer option "--trigger=X" to enable the feature. Possible values for X are:
* 1. Add triggers (if possible) without rewriting subscripts.
* 2. Rewrite complex subscripts, but do not add triggers.
* 3. Rewrite complex subscripts and add triggers.
Note: Has shown positive effect, especially for multi-dimensional arrays. In other cases (especially rewriting complex subscripts) can also cause proofs to fail. So use with care!


sjcjoosten:
Travis seems to pass on everything except `sonar-scanner`. @Sophietje do you know what is going on here?

Sophietje:
> Travis seems to pass on everything except `sonar-scanner`. @Sophietje do you know what is going on here?

My best guess would be that this is due to the fact that this PR is from a fork rather than from a branch on the main repository. See also:

>For security reasons, this advanced feature works only for internal pull requests. In other words, pull requests built from forks won’t be inspected. [[Source](https://docs.travis-ci.com/user/sonarcloud/#analysis-of-internal-pull-requests)]

sjcjoosten:
Thanks for the diagnosis. I guess accepting all pull requests within a branch (after some sanity inspection) should be our default way to handle pull requests from forks then.

PR #244, opened 2019-07-10
fixes #220: add pointer and array support to C once again

### LINKED ISSUE #220: C pointers / arrays
The proposal is to represent both C pointers and arrays as COL arrays:
 - arrays are `Option<Array<T>>`, but may not be assigned `NULL`
 - pointers are `Option<Array<T>>`, but may be assigned `NULL`
 - dereferencing a pointer is subscripting the first element
 - adding to a pointer is taking a slice of the array
 - referencing a name constructs an array that only contains the cell corresponding to the name
 - referencing a subscript also takes a slice of the array

Open issues I see so far:
 - How are we going to make all names everywhere `Cell`, such that they can be referenced?
 - Sometimes we want to "forget" about the size of an array when we are dealing with pointers, how should we do that?

niomaster:


niomaster:
For sonarcloud:
- the bugs should be fixed in #249 
- the coverage is 0.0%, as we don't have a test framework for most portions of vercors yet.

sjcjoosten:
I checked the four SonarCloud issues. The #249 one takes care of all four of them, I'd be happy to ignore the SonarCloud failures in this pull-request.

PR #254, opened 2019-07-12
fixes #243: add reporter that redirects to HRE for viper

### LINKED ISSUE #243: Some messages from Silicon not redirected to HRE
At https://github.com/utwente-fmt/vercors/blob/11df480341c22b6b29ca3db897ec97f2e5319d5f/vercors/viper/src/main/scala/viper/api/SiliconVerifier.scala#L19 `Silicon` is created with a constructor that uses `StdIOReporter()`. We should instead implement a reporter that redirects to to the HRE logging framework and specify it as an explicit reporter.

niomaster:


niomaster:
I was under the impression that viper_api is ours as it bridges vercors with viper, but maybe I'm wrong?

sjcjoosten:
I don't know either. This is why I asked @wytseoortwijn for a review

PR #256, opened 2019-07-16
fixes #230: bug in logging system

### LINKED ISSUE #230: Crash when adding --show-after option
For the category absurd issues, this command works fine:

```vct --silicon --check-history examples/futures/unequalcounting.pvl```

And this crashes vercors:

```vct --silicon --check-history examples/futures/unequalcounting.pvl --show-after silver-class-reduction```

niomaster:


niomaster:
sonarcloud was ignored, because the primary new lines are a new debugging feature that dumps the AST as a tree

PR #260, opened 2019-07-17
Generate triggers for universal quantifiers

sjcjoosten:
In the original pull-request #229, Henk wrote the following:

---

Generate triggers for universal quantifiers.

Rewrite complex subscripts (subscripts that are not allowed in trigger expressions) in quantifiers and (if possible) generate triggers for universal quantifiers (forall and forall*).

Use the Integer option "--trigger=X" to enable the feature. Possible values for X are:
* 1. Add triggers (if possible) without rewriting subscripts.
* 2. Rewrite complex subscripts, but do not add triggers.
* 3. Rewrite complex subscripts and add triggers.
Note: Has shown positive effect, especially for multi-dimensional arrays. In other cases (especially rewriting complex subscripts) can also cause proofs to fail. So use with care!


niomaster:
I believe that we discussed earlier that as the issues I noted are fairly minor and it would be nice that there is a merge point that is just the work of Henk, we should merge this as is.

PR #261, opened 2019-07-17
[feedback wanted] fixes #236: remove contract from parallel regions

### LINKED ISSUE #236: Verification error in case of syntax error
The following program is verified by vercors
```
class C {

  invariant a != null;
  context (\forall* int i; 0 <= i && i < a.length; Perm(a[i], write));
  ensures (\forall int i; 0 <= i && i < a.length; a[i] == 0);

  void allZero(int [] a) {
    // context a != null; 
    par(int i = 0 .. a.length)
       context a != null;
       context Perm(a[i], write);
       ensures a[i] == 0;
    {
       a[i] = 0;
     }
  }
}
```
when the commented out line is un-commented, the verification fails with
```
Errors! (2)
=== exercises/ex3.pvl              ===
   6 
   7   void allZero(int [] a) {
        [------------------
   8     context a != null;
   9     par(int i = 0 .. a.length)
  10        context a != null;
  11        context Perm(a[i], write);
  12        ensures a[i] == 0;
  13     {
  14        a[i] = 0;
  15      }
     ------]
  16   }
  17 }
  CallPreCondition:InsufficientPermission
=== exercises/ex3.pvl              ===
   9     par(int i = 0 .. a.length)
  10        context a != null;
                        [----
  11        context Perm(a[i], write);
                         ----]
  12        ensures a[i] == 0;
  13     {
  caused by
=== exercises/ex3.pvl              ===
   3   invariant a != null;
   4   context (\forall* int i; 0 <= i && i < a.length; Perm(a[i], write));
              [--------------------------------------------------
   5   ensures (\forall int i; 0 <= i && i < a.length; a[i] == 0);
               --------------------------------------------------]
   6 
   7   void allZero(int [] a) {
  PostConditionFailed:AssertionFalse
The final verdict is Fail
```
but it should be a syntax error instead.

niomaster:


niomaster:
I thought there were only a couple of examples that use the contract preceding a parallel block syntax, but it turns out there are quite a few.

niomaster:
Actually, after fixing a bug it turns out to be just ParallelPVLExample1:

https://github.com/utwente-fmt/vercors/blob/master/examples/openmp/parallel-example1.pvl#L11-L35

What should we do about that one?

niomaster:
We only want to keep the thread-specific contracts, so `a != null` etc. should be in the thread contract. This means that `a != null` is auto-quantified, but we want to rely on the jspec rewrite rules to take care of that.

PR #264, opened 2019-07-18
Replace viper modules by hg URLs; Update viper

niomaster:


niomaster:
Will not build because of #269

PR #270, opened 2019-09-06
[feedback wanted] fixes #268: predicate definitions are always final

### LINKED ISSUE #268: Verification aborted exceptionally for simple java code that uses fold/unfold
**This is the code:**

```
public class A{
	int[] a;

	/*@
		resource R(int i) = Perm(a,1) ** true;
	@*/


	/*@ 
		requires R(1);
	@*/
	public void F(){
		//@ refute(false);
		//@ unfold(R(1));
		//@ fold(R(1));
		assert true;
	}
}
```

**This is the output:**

`
Verification aborted exceptionally: java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: java.util.NoSuchElementException: None.get
  Cause: java.util.concurrent.ExecutionException: java.util.NoSuchElementException: None.get
Errors! (1)
[warning] unspecified error exceptional.error
The final verdict is Fail
`

_The problem seems to be related to how the JavaEncode pass is done. An equivalent code in pvl shows no problems at all._

niomaster:
268 has an issue where some indirection is applied to predicates because of inheritance. I went for the easiest solution, which is to say that predicates are always final and cannot be overridden by subclasses, but I would like feedback whether that is the right approach.

### DIFF
--- a/vercors/src/main/java/vct/col/rewrite/JavaEncoder.java
+++ b/vercors/src/main/java/vct/col/rewrite/JavaEncoder.java
@@ -315,6 +315,9 @@ public class JavaEncoder extends AbstractRewriter {
   private boolean is_direct_definition(Method m){
     if (m.isStatic()) return true;
     if (m.isValidFlag(ASTFlags.INLINE) && m.getFlag(ASTFlags.INLINE)) return true;
+    if(m.kind == Kind.Predicate) {
+      return true;
+    }
     if (!m.isValidFlag(ASTFlags.FINAL)){
       m.setFlag(ASTFlags.FINAL, false);
     }


bobismijnnaam:
As a workaround it sounds fine, but if inheritance (as currently is implemented unfinished in vercors) will ever be implemented this behavior will be changed. Relevant (brief) discussion and papers in citations: https://groups.google.com/forum/#!topic/verifast/ubuNHP45TFs .

niomaster:
Thanks for your suggestion Bob. In the meeting we decided that the behavior should be as follows:

- Predicates should not be always final, and we should be able to override them.
- In the case that a subclass overrides a predicate, but not a method that uses the predicate, the method from the parent class should be re-verified using the predicate from the child class.
- The subclass predicate should be stricter than the parent predicate, should we prove that?

bobismijnnaam:
I guess that depends on the use cases? In the case of abstract predicate families the child predicate does not have to be stricter that the parent predicate. In the absence of abstract predicate families it might be good to check if the child is stricter, as it mimics specification inheritance.

PR #273, opened 2019-09-17
fixes #271: Reducible available again, bug in validMatrixFor

### LINKED ISSUE #271: master is broken
Travis doesn't build master

niomaster:
I hope this fixes the master build. @raulmonti I think we should merge this one into master, so that master builds again :)

### DIFF
--- a/vercors/src/main/scala/vct/col/rewrite/RewriteArrayRef.scala
+++ b/vercors/src/main/scala/vct/col/rewrite/RewriteArrayRef.scala
@@ -359,7 +359,7 @@ class RewriteArrayRef(source: ProgramUnit) extends AbstractRewriter(source) {
     var rowJ: ASTNode = create.expression(StandardOperator.Subscript, matrix, quantJ0)
 
     if(seqInfo0.isCell) {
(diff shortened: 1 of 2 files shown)


vercors/src/main/scala/vct/col/rewrite/RewriteArrayRef.scala
```diff
@@ -359,7 +359,7 @@ class RewriteArrayRef(source: ProgramUnit) extends AbstractRewriter(source) {
     var rowJ: ASTNode = create.expression(StandardOperator.Subscript, matrix, quantJ0)
 
     if(seqInfo0.isCell) {
-      conditions += create.starall(quantGuard, create.expression(StandardOperator.Perm, row), quantDecls: _*)
+      conditions += create.starall(quantGuard, create.expression(StandardOperator.Perm, row, create.reserved_name(ASTReserved.ReadPerm)), quantDecls: _*)
```
raulmonti: What is this change doing? I can see you are asking read permissions specifically, but what was it doing before when the read permissions were not specified?

PR #274, opened 2019-09-17
Parallelize build; Add folds to build log (Travis)

niomaster:


niomaster:
I don't understand what code sonarcloud thinks is duplicated...

raulmonti:
I think we should merge this even though SonarCloud complains a little. I do not really understand what it is complaining about, neither Pieter.

PR #276, opened 2019-09-19
Leader election + sequence slicing/updating

wytseoortwijn:


niomaster:
Could you add a header to the new example, so travis can recognize it? Currently the build fails because of it. Thanks! :)

wytseoortwijn:
Just added one

niomaster:
FYI it's now failing on not finding `sonar-scanner`, which has nothing to do with examples...

wytseoortwijn:
Hmmm, but I'm not quite sure if `elect.pvl` still verifies with the latest version of VerCors. If it doesn't, we should exclude it from Travis for now (perhaps I should change the header for this) and create a Git issue for resolving this.

niomaster:
Though I expect #287 to break it

wytseoortwijn:
I also added basic support for sequence slicing (e.g., `xs[i..]`, `xs[..i]` and `xs[i..j]`) and sequence updating (e.g., `xs[i -> 2]`) in PVL.

wytseoortwijn:
Great!

- I'll look into `SequenceUtils` and the CPointers example later. But not today. Probably early next week.
- I think it's a good thing to have a `dev` branch that is separate from `master`. Very nice!
- I see. But I'm not sure if I'll also use branches though. I dislike the idea of using branches instead of forks, because *(1)* it discourages externals (e.g., non-VerCors team members) from contributing to VerCors, and *(2)* it publicly clutters the main repo (we currently seem to have 17 branches already, visible to public, which may or may not close at some point).

wytseoortwijn:
BTW, why is drop used in pointers?

sjcjoosten:
Hi @wytseoortwijn , @niomaster and @raulmonti . Here's a suggestion on how to deal with forks instead of branches:

- encourage external people to use forks
- if it looks any good (if it looks like it may be closed soon), turn it into a branch

This way we get the benefits of both: we can easily do minor edits ourselves, we encourage external people to use forks, and we limit the clutter. Additionally, accepting things into a separate branch tells the contributor "thank you, we plan on using this" a lot quicker, which is a good motivator for them to be more involved in the future too (I've made contributions in the past where the owner said: please fork into a different branch, so I did, then please add documentation to your code, so I did, then I needed to add tests, and at that is the state I left it in about four years ago, that is: I guess I just gave up because learning about his testing framework was just too much work for me, so there's a real risk of not turning forks into branches. Also, people can keep making pull-requests from their fork if they have additional changes once there's a separate branch)

sjcjoosten:
I just followed my own proposal, so at #293 there's a new pull request (for pulling the branch into development)

PR #279, opened 2019-09-24
Webpage

raulmonti:
This merge will only include what I showed in today's VerCors meeting. I wrote down your suggestions and will soon make the modifications for them and a new pull request to introduce them. Since we want the webpage to get updated, I think that this merge should go to master and not dev.


raulmonti:
I am merging the pull request even though some tests were not run on it. Notice that the only files that have changed are concerning to the webpage and thus do not concern to the test suit.

Issue #117, opened 2018-08-23
Workaround fix for `summation-kernel-0` after Viper update


wytseoortwijn:
To fix the `carp/summation-kernel-0.pvl` example after the Viper update (see wytseoortwijn@aeac296), an extra (non-intuitive) lemma was used. The new version of Viper seems somewhat more strict when it comes to matrix multiplication (which is tricky anyway for SMT solvers). As an enhancement, we should try to make this example work without the auxiliary lemma.

wytseoortwijn:
Similar workarounds have been applied in wytseoortwijn/vercors@c0038ed and wytseoortwijn/vercors@02766ca.

wytseoortwijn:
The same fix is applied in `summation-kernel-1.pvl` (e.g. see wytseoortwijn@a34dfd8). One should also fix this example before closing this issue.

niomaster:
Seems like a pretty standard z3 NLA incompleteness thing to me.

PR #284, opened 2019-09-29
fixes #247: array bound checks as precondition of array access

### LINKED ISSUE #247: Clearer indication of missing permissions
It can be very difficult to identify missing permissions when the tool shows an "InsufficientPermission" error.

Take for example the following code:
```
class KernelEncoding {
  given int [] a_in;
  invariant a != null;
  requires a_in != null && a_in.length == a.length;
  context (\forall* int i; 0 <= i && i < a.length; Perm(a_in[i], read));
  context (\forall* int i; 0 <= i && i < a.length; Perm(a[i], write));
  requires (\forall* int i; 0 <= i && i < a.length; a[i] == a_in[i]);
  ensures a[0] == a_in[0];
  void bla (int [] a) {
    a[0] = a_in[0];
  }   
}
```

The verifier will tell you that you have insufficient permissions to verify the ensures clause. However, if you read the context-clauses, then you would expect to have all permissions. After all, you said you require/ensure permission from index i=0 to i=a.length for both arrays.
The issue in this case is however that a.length may be 0 therefore we also need to specify that `invariant a.length > 0`
A more precise error, or possibly a hint about what the issue may be, can alleviate this problem.

niomaster:


niomaster:
The following examples break:

```
arrays/basic-examples.c
arrays/forward-dep-noauto.pvl
arrays/forward-dep.pvl
arrays/JavaArrayExamples
arrays/Transpose.pvl
arrays/vector-add.pvl
carp/forward-loop.c
carp/summation-kernel-0.pvl
carp/summation-kernel-1.pvl
carp/zero-kernel.pvl
carp/zero-loop.c
carp/zero-sub-array.c
carp/zero-sub-matrix.c
openmp/add-spec.c
openmp/addvec1.pvl
openmp/copy-spec.c
openmp/parvec.pvl
openmp/sections-reduced.c
openmp/zero-spec.c
parallel/zero*
quickselect/QuickSelect.java
```

raulmonti:
@Sophietje are you fixing the two java ones ? :D

raulmonti:
Before this branch is merged, we should add a dependency to Mercurial somewhere in the documentation. sbt depends on mercurial and you get a "hg not found" if you do not have it in your computer. I guess the change to sbt was not introduced here but I found it here when trying to compile VerCors to reproduce the tests that fail.

niomaster:
fixes #247

niomaster:
JavaArrayExamples is the one that fails because Z3 can't prove:
```
(  0 < M
&& 0 < N
&& 0 <= i < M
&& 0 <= j < N
) ==> i*N+j < M*N
```

niomaster:
`summation-kernel-*` falls in the same category

wytseoortwijn:
Well, these use non-linear arithmetic. We were essentially just lucky that they passed before.

niomaster:
Certainly! I think this exact problem (squashing a 2d array into a 1d one) is common enough that we should try to prove this type of example though. One solution I tried is using a newer z3 arithmetic solver, which does in fact solve this issue (small exaplanation at the bottom of [this](https://github.com/Z3Prover/z3/releases/tag/z3-4.8.1) release), but that breaks a bunch of other examples (by being too slow). Maybe we could try to cut the length in the presence of a multiplication in a subscript, something like:

```
ar[x*h+y]
-->
loc_2d(ar, x, h, y)
```

where

```
requires (
   (mult1 < len(ar)/mult2 && add1 < mult2)
|| (mult2 < len(ar)/mult1 && add1 < mult1)
|| (mult1*mult2+add1 < len(ar))
loc_2d(ar, mult1, mult2, add1) = VCTArray.loc(ar, mult1*mult2+add1)
```

Of course division is worse in general, but maybe something like w*h/w always simplifies to w? In any case, we would need some way to figure out the "axes" of the array, perhaps just by annotation.

wytseoortwijn:
That probably doesn't actually solve the issue. There is still non-linear arithmetic involved. Could perfectly be that if you give Z3 a new random seed for example, it doesn't verify any longer. Even changing a variable name might slightly change the search heuristics of Z3, causing verification to fail.

wytseoortwijn:
One thing you could do indeed is hide multiplication inside a function or predicate definition, like your suggestion of `loc_2d`. That might work perhaps...

niomaster:
Just connecting some organizational dots here:

 * I believe the lemma in #117 solves the same issue I noted, so we should solve those together
 * I think the `RecognizeMultiDim` pass does the opposite of what we want (int[N][M] -> int[N*M]); perhaps we should remove that or find an alternative if it serves a purpose.

niomaster:
array bounds checking is included in #629

PR #289, opened 2019-10-01
fixes #288: Improve SIL for assignments and returns of method invocations

### LINKED ISSUE #288: Improve SIL for assignments and returns of method invocations
Generated SIL code for method invocations is a bit verbose. The following PVL code:

```
class MyClass {
  boolean baz() {
    return true;
  }

  boolean foo() {
    boolean e = baz();
    return baz();     
  }
}
```

Gets transformed into the following SIL:

```
method method_MyClass_foo(diz: Ref, current_thread_id: Int, globals: Ref) returns (sys__result: Bool)
  requires diz != null
  requires 0 <= current_thread_id
{
  var e__1: Bool
  var __flatten_1__2: Bool
  var __flatten_2__3: Bool
  var __flatten_3__4: Bool
  // refute
  assert !false
  __flatten_1__2 := method_MyClass_baz(diz, current_thread_id, globals)
  e__1 := __flatten_1__2
  __flatten_3__4 := method_MyClass_baz(diz, current_thread_id, globals)
  __flatten_2__3 := __flatten_3__4
  sys__result := __flatten_2__3
  // assert
  assert true
  inhale false
}
```

It would be nice for readability for the output SIL if the flatten variables were not there, and instead the results would be assigned directly to the variables. I would like to work on this.

bobismijnnaam:
Changed the flatten and voidcalls passes to not flatten the first method invocation encountered. Fixes #288.


vercors/src/main/java/vct/col/rewrite/VoidCalls.java
```diff
@@ -104,31 +113,44 @@ public void visit(MethodInvokation e){
...
-        result=res;
-        return;
-      }
     }
     super.visit(s);
   }
+
+  /**
+   * Turns a method invokation with a location into a method invokation that assigns the result to the first parameter.
+   * Only works for plain methods, throws otherwise. Applies rewrite to both the method invokation and its constituents, and the location.
```
raulmonti: Thanks Bob for commenting!! :D

bobismijnnaam:
The issue #288 is a bit unclear. For the following code snippet:

```
class MyClass {
  boolean baz() {
    return true;
  }

  boolean foo() {
    boolean e = baz();
    return baz();     
  }
}
```

foo() is now flattened into:

```
method method_MyClass_foo(diz: Ref, current_thread_id: Int, globals: Ref) returns (sys__result: Bool)
  requires diz != null
  requires 0 <= current_thread_id
{
  var e__1: Bool
  // refute
  assert !false
  e__1 := method_MyClass_baz(diz, current_thread_id, globals)
  sys__result := method_MyClass_baz(diz, current_thread_id, globals)
  // assert
  assert true
  inhale false
}
```

I am not sure if I can add a testcase for this. If not then the code can be merged as far as I'm concerned.

raulmonti:
Sonnar is not running because Bob is external. Since all other tests pass I will merge his contribution now.

PR #290, opened 2019-10-01
add pointer_index, equivalent of pointer for one element at an index

niomaster:


raulmonti:
We should look at sonarcloud suggestions in the future and see if we really want to follow them. Otherwise it is just disturbing and we never really follow them.

PR #292, opened 2019-10-06
Adding SplitVerify

sjcjoosten:
Hi all, please comment on whether or not this should be a separate repo. Here is the source to SplitVerify, and I hope you find it useful (and also I hope you find it useful that it would reside here in this repo).


raulmonti:
I think it fits very well here in the same repo as VerCors 👍

PR #311, opened 2019-10-23
Clean up repository; Remove depdency on ant and module system.

### LINKED ISSUE #93: Building on Windows
VerCors does not seem to build under Windows (with e.g. Cygwin).

### LINKED ISSUE #197: Old sbt version blocks java >8
The current version of `sbt` that we distribute has a bug where it cannot deal with modern java versions, a fresh install of sbt works fine. I question though whether we should distribute sbt at all, as it is already required to install in the build instructions anyway.

### LINKED ISSUE #198: Remove build.xml
Now that all compilation logic is moved to `sbt`, we should try to move the remaining `ant` tasks to `sbt` as well. Specifically, `javadoc` and `dist` are still handled by `ant`.

### LINKED ISSUE #218: Installation - java8 cask no longer available
The installation instructions for the requirements of VerCors include the following command: `brew cask install java8`
However, the cask java8 is no longer available (due to Oracle license changes?), therefore we need to update the installation instructions.

A solution can be found here: https://stackoverflow.com/questions/24342886/how-to-install-java-8-on-mac/55774255#55774255

### LINKED ISSUE #255: Non-existing VCT HOME value results in no output
When setting up my IntelliJ such that I could run VerCors from within the IDE, I used the value `"."` instead of `.` for the VCT_HOME environment variable. However, there should not be any quotation marks. As a result, VCT_HOME cannot be found and an exception is thrown (when running in debug mode): `java.lang.Error: VCT_HOME value "." is not a directory`.
This exception should be handled such that the user is nicely informed about the issue.

### LINKED ISSUE #265: Build successful after unsuccessful build
I got the following output from running `ant`:

```
Buildfile: /Users/sjc/Documents/Twente/VerCors/vercors/build.xml

compile:
     [exec] [info] Loading settings for project global-plugins from idea.sbt ...
     [exec] [info] Loading global plugins from /Users/sjc/.sbt/1.0/plugins
     [exec] [info] Loading project definition from /Users/sjc/Documents/Twente/VerCors/vercors/vercors/project
     [exec] /Users/sjc/Documents/Twente/VerCors/vercors/vercors/build.sbt:34: error: not found: value assembly
     [exec]     assembly / mainClass := Some("vct.main.Main"),    // Define JAR's entry point
     [exec]     ^
     [exec] /Users/sjc/Documents/Twente/VerCors/vercors/vercors/build.sbt:35: error: not found: value assemblyMergeStrategy
     [exec]     assemblyMergeStrategy in assembly := {
     [exec]     ^
     [exec] /Users/sjc/Documents/Twente/VerCors/vercors/vercors/build.sbt:35: error: not found: value assembly
     [exec]     assemblyMergeStrategy in assembly := {
     [exec]                              ^
     [exec] /Users/sjc/Documents/Twente/VerCors/vercors/vercors/build.sbt:36: error: not found: value MergeStrategy
     [exec]       case "logback.xml" => MergeStrategy.first
     [exec]                             ^
     [exec] /Users/sjc/Documents/Twente/VerCors/vercors/vercors/build.sbt:38: error: not found: value assemblyMergeStrategy
     [exec]         val oldStrategy = (assemblyMergeStrategy in assembly).value
     [exec]                            ^
     [exec] /Users/sjc/Documents/Twente/VerCors/vercors/vercors/build.sbt:38: error: not found: value assembly
     [exec]         val oldStrategy = (assemblyMergeStrategy in assembly).value
     [exec]                                                     ^
     [exec] [error] sbt.compiler.EvalException: Type error in expression
     [exec] [error] sbt.compiler.EvalException: Type error in expression
     [exec] [error] Use 'last' for the full log.
     [exec] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? 
     [exec] Result: 1

BUILD SUCCESSFUL
Total time: 12 seconds
```

This issue is there to complain about the final `BUILD SUCCESSFUL` message, which should not be there if the project is not successfully built (as is the case).

### LINKED ISSUE #304: Scripts in unix/bin, windows/bin broken
The directories in the title contain scripts to run vercors, which work great for setting JVM parameters, setting up environment variables etc. There's also a collection of other scripts which are either quite useful, but no longer work, or are obsolete. We should look into either removing or fixing the scripts as necessary.

niomaster:
The larger idea of this pull request is to unify the repository under one build tool, the scala build tool. As it turns out, this required me to tear out quite a few interconnected components. As always, I tried to use my best judgement, but please let me know if you disagree with a decision I made. I see the PR as a `draft' for the best way forward.

* Removed /deps/ant, as I think we should not distribute it ourselves, and we no longer depend on ant.
* Removed build.xml from vercors and subprojects
* Moved /unix/bin/* and /windows/bin/* into bin
* Cleaned up dependencies, which included things as extra header files and examples (mostly z3)
* No longer distribute commons-lang, sbt should take care of all library dependencies automatically
* Remove src/overview.html
* Removed the modules system from VerCors, as I think it's overkill for our case.
* Transitioned from `sbt-pack` and `sbt-assembly` to `sbt-native-packager`
* Moved /deps, /config, /include and /examples/selftest to a resource directory
* Moved /vercors/* to /, so the root of the repository is the root of the build.

This PR fixes #198, fixes #197, fixes #304, fixes #255, fixes #265, fixes #218 and fixes #93


niomaster:
*!!* Before mering we should take a look at the LICENSE file. It currently states the repo is MPL, excluding `deps`, `viper` and `examples`. I think the correct changes to this are:

* Change deps to the new path `vercors/src/main/universal/res/deps`
* Remove `viper`, as that source is now just `viper_api`, which is ours. The viper sources are only referred to via the repository.
* `examples` should probably stay as is, as it seemes to be licensed CC0

Furthermore, what used to be the vercors subdirectory also contained a copy of the MPL, but as it is now in the same directory as the root that was removed. Also, hre contains a copy of the MPL separately?

raulmonti:
> * CC0

I think your suggestions are fine. Does anybody else have something to say? Otherwise I like very much this cleanup and we should merge it.

PR #314, opened 2019-10-28
Add conditions for (z)frac; cast integers to fractions automatically

### LINKED ISSUE #277: Fractions compile to integer and you get type error
The following example is not working right:
```
given frac p;
invariant 1/5 < p && p < 1/2;
```
The important part of the error message is:
```
[debug] At Main.java:634: 
[debug] java.lang.AssertionError: assertion failed: Expected both operands to be of the same sort, but found Int (1 / 5) and Perm (p@12@03).
[debug] At Main.java:634: 
[debug] 	at scala.Predef$.assert(Predef.scala:219)
[debug] At Main.java:634: 
[debug] 	at viper.silicon.state.terms.Less.<init>(Terms.scala:845)
[debug] At Main.java:634: 
[debug] 	at viper.silicon.state.terms.Less$.apply(Terms.scala:854)
[debug] At Main.java:634: 
[debug] 	at viper.silicon.state.terms.Less$.apply(Terms.scala:850)
```
Furthermore, when I use --compiled option I find that it translates the second line to:
`invariant 1 \ 5 < p && p < 1 \ 2
`
It writes `\` instead of `/`. We may need to check if we are doing something wrong there or if it is viper's printing what is going wrong.
One more thing that I found inconvenient is that when we write:
```
given frac p;
invariant 0/1 < p && p < 1/1;
```
we get ` 0 < p && p < 1`  in the compiled file and again VerCors complains about type checking:
```
[debug] At Main.java:634: 
[debug] java.lang.AssertionError: assertion failed: Expected both operands to be of the same sort, but found Int (0) and Perm (p@12@03).
[debug] At Main.java:634: 
```

niomaster:
fixes #277


niomaster:
please note I decided to go with \ for fraction division for now, as f/ causes parsing issues in java, e.g. "1f/2" is equivalent to divide the float 1.0 by the integer 2.

PR #318, opened 2019-10-28
flatten if statements and their conditions

niomaster:


raulmonti:
I will merge this even tough SonarCloud complains about test coverage of new code. Actually examples/basic/InvokationInGuard.java would be a kind of test covering the new code, but we need to find out how to tell SonarCloud to take this into consideration.

PR #319, opened 2019-10-29
clean up PVL expression grammar

niomaster:
This change is still very restrictive in what a left-expr may be, but at least improves the situation for regular expressions.


niomaster:
I would like to think a little bit more about how to handle lexprs (here named target) before we merge.

Issue #322, opened 2019-10-30
Installation instructions are outdated


niomaster:
The installation instructions on the website are outdated: <https://utwente-fmt.github.io/vercors/Installation.html#install>. They should either reflect the instructions in the main README, or refer to the readme.

niomaster:
The website is now out of scope for this repository.

Issue #320, opened 2019-10-29
Installation instructions and vct file


sjcjoosten:
The releases work great! It would be nice if the releases were to include:

- a shell file that calls java with the usual parameters
- a readme file that tells you to use the shell file
- a version number

It would also be nice if the instructions to use a release were referenced in the installation instructions.

Here's a proposal for a shell file (after the other vct file, also returns 1.1.0 as a version, alternatively we might wish to put the version number in a separate file and just echo that file):

```
#!/bin/bash

# find out location of binary
DETECT=`dirname "$0"`
# parent is platform directory
DETECT=`dirname "$DETECT"`
# parent is home
export VCT_HOME=`dirname "$DETECT"`

function f() {
  echo $VCT_HOME | sed 's#^/\([a-z]\)\(/\|$\)#\1:\2#' | sed 's|/|\\|g'
}

case `uname -s` in
MINGW*)
SEP=";"
VCT_HOME=`f`
;;
*)
SEP=":"
;;
esac

if [ -d $VCT_HOME/deps/modules ] ; then
  TOOL_HOME=$VCT_HOME/deps
else
  TOOL_HOME=`dirname $VCT_HOME`
  TOOL_HOME=`dirname $TOOL_HOME`
fi
export TOOL_HOME

VCT_PATH=$VCT_HOME/vct-tool.jar

VMARGS="-Xss128M"
i=0
args=()
for arg in "$@" ; do
  case $arg in
    -X*)
      VMARGS="$VMARGS $arg"
    ;;
    --version)
      echo "vct version 1.1.0"
      exit 0
    ;;
    *)
      args[$i]="$arg"
      ((++i))
    ;;
  esac
done

exec java $VMARGS -cp "$VCT_PATH" vct.main.Main $@
```

Edit: I needed to get an exit-code of 0 instead of 1.

niomaster:
Perhaps you missed it, but with #311 this has been partially solved. We now use `sbt-native-packager`, which amongst other things generates releases with run scripts for us. Currently we just map the main `README.md` to the release, but that might indeed not be very applicable to how you run vercors in the release. Re: version numbers, indeed we should have a way to output them. I would however like to find a way to only have the version number in one place, preferrably build.sbt.

sjcjoosten:
Sorry for inflating so many issues in one place. Here's what's left:

- version numbers
- get instructions on how to use the release inside the release itself
- instructions on how to download and use the release on the VerCors-related websites.

[...]

PR #323, opened 2019-10-31
Release 1.2.0

niomaster:
I would like to do a release soon, as I need a public .deb build for the docker image I'm building.


raulmonti:
@niomaster Can we use this opportunity to review the readme? specifically the installation instructions?

niomaster:
The installation instructions are correct for building from the git repository, but you're probably referring to #322 or #320?

raulmonti:
Oh, good, but we can fix those two next time. I will review this now.

build.sbt
```diff
@@ -0,0 +1,70 @@
...
+lazy val viper_api = (project in file("viper"))
+lazy val parsers = (project in file("parsers"))
+
+lazy val vercors = (project in file("."))
+    .dependsOn(viper_api)
+    .dependsOn(parsers)
+    .settings(
+        name := "Vercors",
+        organization := "University of Twente",
+        version := "1.1.0",
```
raulmonti: Do we need to change this to 1.2.0?

niomaster: Good catch, yes.

raulmonti: I guess this small change does not deserve a full travis test. So I will merge now.

PR #327, opened 2019-11-04
Print warning when sonar is not executed in CI check

bobismijnnaam:
Ensure a warning is printed if sonar will fail because the environment is not secure.

Info:
https://travis-ci.community/t/pull-request-builds-fail-with-sonar/3473
https://docs.travis-ci.com/user/environment-variables/ (Look for `TRAVIS_SECURE_ENV_VARS`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ before_cache:
   - rm -fv $HOME/.ivy2/.sbt.ivy.lock
   - find $HOME/.ivy2/cache -name "ivydata-*.properties" -print -delete
   - find $HOME/.sbt        -name "*.lock"               -print -delete
+  - rm -rf $HOME/.sbt/boot
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,4 +45,10 @@ matrix:
       - SPLIT=4/5 ./.travis/build.sh
     - script:
       - ./.travis/travis_fold.sh build "Build Vercors" "sbt compile"
+      - echo "TRAVIS_SECURE_ENV_VARS=${TRAVIS_SECURE_ENV_VARS}";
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,6 +30,7 @@ before_cache:
   - rm -fv $HOME/.ivy2/.sbt.ivy.lock
   - find $HOME/.ivy2/cache -name "ivydata-*.properties" -print -delete
   - find $HOME/.sbt        -name "*.lock"               -print -delete
+  - rm -rf $HOME/.sbt/boot  # Needed to fix the "nsc.Global not found" error.
```
niomaster: So, the core problem with the build failing is that SBT depends on a version of scala (in this case 2.12.8) that has somehow become corrupted in the cache. There is no other way to solve this than to remove the broken scala from the cache somehow, as SBT has no way to bootstrap other than to use a working scala.

This fixes the problem in the most fail-safe way, namely removing the scala versions installed if it is cached. The disadvantage is that SBT has to pull all dependencies for each build, and viper has to be rebuilt each time as well (as the scala install changed). I think as we've seen this issue only one time now, this solution is too heavy-handed, but a good option for stability if we see it more often.

Let's for now just remove the caches currently in travis, and see if we observe the problem again. Perhaps we'll gain some useful information from that as well.

Otherwise, no problem with the pull request; great addition!

bobismijnnaam: Okay! Removed the `rm` per 4b664e2c704e16bfa7bb4ad0a4c2cb719183c0f9.

raulmonti:
Sonnar cloud fails because Bob is an external contributor in this PR. This is nice because you can see his error message being displayed :D. I will merge this since everything else looks good.

PR #331, opened 2019-11-06
fixes #308: Error on unsatisfiable method preconditions

### LINKED ISSUE #308: Unsatisfiable requires, possible warning?
Currently the following code verifies:

```
final class MyClass {
    public static int foo;

    //@ requires Perm(MyClass.foo, 1);
    public static void eat() {}

    //@ requires Perm(MyClass.foo, 1) ** Perm(MyClass.foo, 1);
    //@ ensures Perm(MyClass.foo, 1) ** Perm(MyClass.foo, 1);
    public static void main() {
        foo = 2;
        eat();
        eat();
        //@ assert 1 == 2;
    }
}
```

I think it boils down to the fact that the requires in this case is false indirectly. In the most cases I expect at least a warning (```warning: requires might be unintentionally unsatisfiable```), at most an error (```error: requires must be satisfiable```). 

The only cases that I can come up with where ```requires false;``` could be useful is in the context of inheritance, or if you want to exclude a method from verification. E.g. ```ImmutableArrayList extends ArrayList```, and then give ```add()``` a ```requires false;``` to indicate it must never be called when casted. The second case can actually be useful if you want to have some "dirty" code in your class. However, in both cases I think a warning would still be okay, or maybe an annotation can be required to suppress the error (since these cases are the exception, so it should not happen too often anyway).

In theory checking satisfiability in something like Z3 is pretty standard, but I don't know how to do it in Viper.

bobismijnnaam:
This pull request repairs satisfiability checking for method preconditions. I also cleaned up SilverImplementation.scala (results were printed there, as well as in SilverBackend.java. Moved it all to SilverBackend.java) and ViperAPI.java (this api claimed to do reachability checking, but it never did such thing - it only copied a few empty arrays. Removed this parameter to signify absence of this feature in the backend). Lastly I added vim swap files to the gitignore list, to prevent me from accidentally committing them.

It can be turned off by supplying the flag `--disable-sat`.

### DIFF
--- a/src/main/java/vct/col/rewrite/SatCheckRewriter.java
+++ b/src/main/java/vct/col/rewrite/SatCheckRewriter.java
@@ -17,11 +17,6 @@ import vct.col.ast.stmt.decl.ProgramUnit;
 import vct.col.ast.type.Type;
 import vct.col.util.OriginWrapper;
 
-// Encountered bugs:
-// TODO: (In SilverBackend.java) This bad boy swallows exceptions without a sound?
-// TODO: Contract.getOrigin() is null below?
-// TODO: pass/fail is not adhered to in test?
-
 // Still need to do:
 // TODO: Java tests
 
--- a/src/main/java/vct/col/rewrite/SatCheckRewriter.java
+++ b/src/main/java/vct/col/rewrite/SatCheckRewriter.java
@@ -17,9 +17,6 @@ import vct.col.ast.stmt.decl.ProgramUnit;
 import vct.col.ast.type.Type;
 import vct.col.util.OriginWrapper;
 
-// Still need to do:
-// TODO: Java tests
-
 public class SatCheckRewriter extends AbstractRewriter {
     public static class AssertOrigin extends BranchOrigin {
         public AssertOrigin(String branch, Origin base){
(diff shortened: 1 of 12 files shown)


src/main/java/vct/col/rewrite/SatCheckRewriter.java
```diff
@@ -13,43 +18,64 @@
...
+                    m.getKind(),
+                    rewrite(m.getReturnType()),
+                    rewrite(m.getContract()),
+                    "__contract_unsatisfiable__" + m.name(),
+                    rewrite(m.getArgs()),
+                    m.usesVarArgs(),
+                    blockStatement
+            );
+
+            blockStatement.setParent(assert_method);
```
niomaster: Was this necessary? Just curious.

niomaster:
The travis failure is due to the sonar job erroring again

src/main/java/vct/col/rewrite/SatCheckRewriter.java
bobismijnnaam: Using a block statement was, but the setparent wasn't. I (randomly) noticed the field while I was debugging and thought it might be good to do. But if it isn't done consistently anywhere maybe line 75 should be removed (and possibly the parent field as well, if it isn't used anymore)

PR #333, opened 2019-11-07
Add type-checking to jspec rules

### LINKED ISSUE #309: simplification failure for pointer index
If you have a `\pointer_index` in a `par` block, we want a simplification to occur that does not. If we take this example:

```
par(int i = 0; i < len; i++)
  context \pointer_index(array, i, write);
{ ... }
```

First `\pointer_index` is desugared:

```
par(int i = 0; i < len; i++)
  context array != null ** i <= array.length ** Perm(array[i], write);
{ ... }
```

Then the parallel blocks are encoded, by quantifying over all threads:

```
context forall* (0 <= i < len): array != null ** i <= array.length ** Perm(array[i], write)
method par_body() {...}
```

Then, somewhere in a simplifying step two things happen: the conjunctions are split out over different quantifiers:

```
context forall* (0 <= i < len): array != null
context forall* (0 <= i < len): i <= array.length
context forall* (0 <= i < len): Perm(array[i], write)
method par_body() {...}
```

There is also a rule that removes the quantifier from quantified expression of which the body is independent of the quantified variable, where the body is not a resource. We want this to happen for `array != null`, but it does not. If in the initial example you additionally specify `array != null`, this is done, so either the stumbling block is that the quantifier remains a forall-star, even though the body is a boolean, or the rule ordering causes this simplification to fail.

niomaster:
This fixes #309. I've added type-checking to the rewrite rules (simplify_quant etc.). The rules used to be applied structurally without checking type, but now whenever a generic expression occurs, the true AST must be typed, and be a subtype of the denoted type in the rule. This now admits a rule:

```
axiom starall_bool {
  (\forall* int i; b1; b2)
    ==
  (\forall int i; b1; b2)
}
```

where b1, b2 are booleans, which solves the \pointer_index simplification. I've made rules more general that were no longer applied due to the type being too strict, but there's still plenty of (non-applied) low-hanging fruit there I think.

Lastly, I'd like to draw attention to a change I've made in RewriteSystem, i'll comment once I can view the changes...

### DIFF
--- a/examples/basic/par-pointer-simplification.c
+++ b/examples/basic/par-pointer-simplification.c
@@ -1,0 +1,11 @@
+//:: cases ParPointerSimplification
+//:: verdict Pass
+
+//@ context \pointer(ar, len, write);
+void test(int ar[], int len) {
+    for(int i = 0; i < len; i++)
+    //@ context \pointer_index(ar, i, write);
+    {
+
+    }
+}
--- a/parsers/src/main/antlr4/Java7JML.g4
+++ b/parsers/src/main/antlr4/Java7JML.g4
@@ -6,7 +6,7 @@ identifier : javaIdentifier ;
 
 extraIdentifier : valReserved ;
 
-extraType : 'resource' | 'process' | 'frac' | 'zfrac' ;
+extraType : 'resource' | 'process' | 'frac' | 'zfrac' | 'rational' ;
 
 extraAnnotation
     : 'pure'
--- a/src/main/java/vct/col/ast/type/PrimitiveType.java
+++ b/src/main/java/vct/col/ast/type/PrimitiveType.java
@@ -306,6 +306,8 @@ public final class PrimitiveType extends Type {
       return new ConstantExpression(0);
     case Fraction:
       return new ConstantExpression(0);
+    case Rational:
+      return new NameExpression(ASTReserved.NoPerm);
     case Integer:
       return new ConstantExpression(0);
     case Long:
--- a/src/main/java/vct/col/rewrite/RewriteSystem.java
+++ b/src/main/java/vct/col/rewrite/RewriteSystem.java
@@ -1,7 +1,5 @@
 package vct.col.rewrite;
 
-import static hre.lang.System.Debug;
-import static hre.lang.System.Fail;
 import hre.ast.Origin;
 import hre.lang.HREError;
 import hre.lang.Ref;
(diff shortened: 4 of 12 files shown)


niomaster:
It's about the change to this method here: 

https://github.com/utwente-fmt/vercors/blob/issue-309/src/main/java/vct/col/rewrite/RewriteSystem.java#L441-L468

Where the change is this:

https://github.com/utwente-fmt/vercors/pull/333/files#diff-353a3308331f8a406cdce11777608772R459

In this context we are rewriting the right-hand side of a rule ("axiom"), e.g. `axiom { e1 + e1 == 2*e1 }`: we would now be rewriting `2*e1`. We rewrite any names matched in the left hand side in the right hand side, so in this case `e1` would be rewritten to whatever expression it matched. This change is about binding expressions (forall(*), let, etc.) and concern the "init" of a declaration in it, which is mostly applicable to let expressions:

```
\let int i = 3; (i*i)
```

Here `3` is the init. I believe that in the first part of the linked method, the init expressions are all rewritten. However, when the binder is of type `Let`, as in the changed line, the expressions are rewritten _again_. I think this can lead to incorrect mappings. The case I found was from this rule:

```
axiom simplify_shift_1 {
    (\forall* int i; b1 ; Perm( ar [ (e1!i) + i ] , (e2!i) ))
      ==
    (\forall* int k_fresh ; (\let int i=k_fresh-e1;b1) ; Perm( ar [ k_fresh ] , e2 ))
  }
```

Matching on something like:

```
\forall* int j; b1; Perm(ar[i*N + j], write)
```

Note that `i` is free in this expression (it is bound in a parent expression). Our mappings then are `e1 -> i*N`, `e2 -> write`, `i -> j` and some other expression for `b1`. In the right-hand side, `\let` is a binder, so we rewrite the init part with the mappings to `k_fresh_rw_0 - i*N` (k_fresh is made unique, `e1` is known). Because specifically this is a `\let`, the expression is rewritten again to `k_fresh_rw_0 - j*N` (As `i` is still mapped to `j`). This is clearly incorrect.

I'm asking because I want to be sure I've really understood this correctly. The rules contain a lot of rules with this structure, so I'd think we'd have noticed this before, but maybe I'm wrong.

raulmonti:
@wytseoortwijn can you check this if you have some time? I believe the original code is yours, maybe you have a deeper understanding about it? With Pieter we think that what he is doing is OK but maybe we are missing something ...

niomaster:
My mistake, I misread the blame! The code in question is actually Stefan's

PR #335, opened 2019-11-08
fixes #332: contract origin null

### LINKED ISSUE #332: Contract origin is null early on in verification
I have noticed that the origin of the contract is null very early on in the verification process. Not sure how to make a test case for this, but I have a screenshot of a debug print in `java_resolve`:

image

Already in the java_resolve phase it is null, even though the contracts themselves are correct. I suppose the contracts should have origins, just like other ast nodes. I would like to look into this.

bobismijnnaam:
This fixes #332 by:

- making sure when contracts are rewritten the origin is preserved (which happens in quite a few places)
- when contracts are collected by SpecificationCollector, the individual regions from the ASTNodes are combined into one region when contract parts are collected into a contract

I thought about making ContractBuilder aware of origins, as this was mostly caused by that. However, this would result in an incomplete view of the complete origin on the contract. Since given a `requires P` clause, a ContractBuilder is only given the `P` part of the requires clause, not the whole `requires P` ASTNode. Hence, if ContractBuilder would track the origins, the `requires` part would disappear from the origin, which is ugly.

Also, it seems not something ContractBuilder should concern itself with - it seems like more of a parser problem. The approach added to SpecificationCollector seems to be in harmony with how SpecificationCollector works at the moment.

### DIFF
--- a/examples/basic/ContractSatisfiable.java
+++ b/examples/basic/ContractSatisfiable.java
@@ -34,7 +34,7 @@ class MyClass {
   }
 
   // User error, should be detected
-  requires a > 0 && a < 0 && b > 10;
+  //@ requires a > 0 && a < 0 && b > 10;
   void complicated_arguments(int a, int b) {
     // Should not be detected
     //@ assert b < 10;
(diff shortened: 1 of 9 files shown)


raulmonti:
Merging, since all test pass but sonnar cloud.

PR #338, opened 2019-11-12
Fix unintended "possible simplification failure"

niomaster:


niomaster:
By the way, I really dislike comparing hard-coded names like I'm doing here, but that's one for the milestone "fix all naming collisions" :)

PR #350, opened 2019-12-09
Adding functionality to sequences in PVL

OmerSakar:
This PR adds the following functionality on sequences to PVL:

- Appending and prepending a single value to a sequence by using the operators `::` and `++` respectively.
- A function `removeAt(xs, i)` to remove index `i` from a sequence `xs`.
- A function `isEmpty(xs)` which checks if a sequence, set or bag is empty.
- Simpler constructors for sequences, sets, and bags. The old constructors are still present. For sequences, the syntax is `[1, 2, 3]` or if the sequence is empty `[t:int]`. For sets, the syntax is `{1, 2, 3}` or if the set is empty `{t:int}`. For bags, the syntax is `b{1, 2, 3}` or if the bag is empty `b{t:int}`.


raulmonti:
Nice!!!! :D

PR #361, opened 2020-01-06
Gpgpu support

niomaster:


niomaster:
The above commit adds `\min` and `\max` binders. Importantly, they are not translated into viper yet (because I don't think that's built in), but we blindly hope that they are simplified away. Perhaps we want to support them similarly to what @OmerSakar is working on; i.e. via axiomatization in VectorExpression.

niomaster:
I've though a little bit more about it and while we could write an axiomatization for max/min, I don't think it would ever be useful. `max` is simplified in any expression using combinations of `*`, `+`, `-` and unary `-`, where the quantified variable (say `i`) does not occur in multiple branches in the expression tree. This means that for simplification to fail, either `i` occurs on both sides of a binary expression, or the expression contains division. I strongly doubt the solver would produce an answer for such an expression.

What I've done now is to report an error whenever min/max is not simplified away asking to report it.

niomaster:
I've reformatted the jspec rules, so for the careful reader, these have been added:

* negation_inwards_add
* negation_inwards_sub
* negation_inwards_mul
* negation_inwards_div
* simplify_bound
* simplify_bound_eq
* simplify_bound_max
* simplify_bound_eq_max
* simplify_bound_max_right
* simplify_bound_eq_max_right
* max_indep_add
* max_indep_add_right
* max_indep_sub
* max_indep_sub_right
* max_indep_mult
* max_indep_mult_right
* max_quant_var
* max_neg_quant_var

niomaster:
Seems I was wrong again, for example a sortedness condition from a bubblesort algorithm:

```
loop_invariant (\forall int k; 0 <= k && k < i; out[k] <= out[i]);
```

gets "simplified" to:

```
(\max int k;0 <= k && k < i;out[k]) <= out [i]
```

So in this case it would have been much better to keep the original condition. Perhaps the right approach is to try to simplify the max expression, but if it fails, revert to the `\forall` expression. That would require a rewrite pass instead of the rewrite rules.

Issue #360, opened 2019-12-19
Matching loop in the axiomatization of VectorExpressions


OmerSakar:
I looked into the summation of a sequence of integers which is supported in PVL (possibly also different frontends).

I found that `(\sum int k; 0 <= k && k < |xs|; xs[k])` or `\sum([0 .. |xs|),xs)` is the way to sum a sequence of integers `xs`. This is translated into Viper as a VectorExpression which is fine, however reasoning about the sum of a sequence seems to not work. For example, the PVL method below will not verify with the current axioms:

```
    seq<int> xs = seq<int> {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
    assert (\sum int k; 0 <= k && k < ; xs[k]) == 55;
```
The Viper output of the relevant part (without the flatten variables) is:
```
  var xs: Seq[Int]
  xs := Seq(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
  assert vsum(vrange(0, |xs|), vseq(xs)) == 55
```
 
I have looked into the Viper output and put it through a program Axiom Profiler (made by ETH Zürich, https://bitbucket.org/viperproject/axiom-profiler/src/default/) and I noticed the following problem. One of the axioms is not properly defined. 

The problematic axiom is the following (and can be found in the `VectorExpression` domain):

```forall xs : VectorExpression[T], i:Int, j:Int :: { vsum(vrange(i,j),xs) } vsum(vrange(i,j+1),xs) == vadd(vsum(vrange(i,j),xs),vget(xs,j))```

In the example above the following will happen, firstly `|xs| == 10` will be asserted. Using this fact, the assertion is rewritten as `vsum(vrange(0, 10), vseq(xs)) == 55`. Lets focus on the left side of the comparison.

The trigger of the axiom matches the left side. `i` will be bound to 0 and `j` to 10 and thus the following term will be added.

```
vsum(vrange(0, 10+1), vseq(xs)) == vadd(vsum(vrange(0, 10), vseq(xs)), vget(vseq(xs), 10))
```

The trigger matches the left side again where `i` is bound to 0 and `j` is bound to 11. As you might already expect, there is a matching loop. I think that, due to this matching loop, we cannot reason about the summation a sequence of integers.

OmerSakar:
[...]

PR #364, opened 2020-01-07
Fixes issue #360: Allow reasoning on sums of sequences of ints.

OmerSakar:
Changed the axiom as described in a comment on the issue ( #360 ). Added a simple example to show a case with concrete values and with no assumption on the values.


raulmonti:
solves #360

PR #372, opened 2020-01-27
LaTeX listings syntax highlighting for PVL code 

OmerSakar:
I wrote a LaTeX listings language for PVL code. It is a package named listings-pvl.sty and can be imported using the usepackage command. I have also written a small document (named documentation.tex) to provide some information on how to use it and how to change it. 

The package and the documentation have been written with two things in mind: extensibility by the user and easily changeble when PVL itself changes.

The exact coloring scheme is up to debate. See documentation if you want to change it for your own project or change the package itself.

My question is whether we want this in the VerCors repository or not.

To get an idea of how this would be used, a simple document would be:

```
\documentclass[]{scrartcl}
\usepackage[T1]{fontenc} % Optional if encoding breaks

\usepackage{listings-pvl} % Defines the PVL language

\begin{document}
\begin{lstlisting}[language=PVL]
class Incrementer {
    requires n >= 0;
    ensures |\result| == |input|;
    ensures (\forall int k; 0 <= k && k < |\result|; \result[k] == input[k]+n);
    seq<int> incrementAllByN(seq<int> input, int n) {
        seq<int> res = seq<int> {};
        int i = 0;
        
        loop_invariant 0 <= i && i <= |input|;
        loop_invariant i == |res|;
        loop_invariant (\forall int k; 0 <= k && k < i; res[k] == input[k]+n);
        while (i < |input|) {
            res = res + seq<int> {input[i]+n};
            i = i + 1;
        }
        return res;
    }
}
\end{lstlisting}
\end{document}
```

With as result:
this_one


niomaster:
Perhaps you could look into also publishing this to CTAN?

OmerSakar:
That is a good idea, I will upload it to CTAN.

PR #378, opened 2020-02-03
Bump Z3 version

### LINKED ISSUE #375: Example demo/demo4.pvl is broken because of current dependencies
`demo/demo4.pvl` currently does not verify in vercors 5dfa950cd8e1e7c457d45c24dd4819ca64de1a53. In vscode (don't know how to see the version?) it verifies most of the file, except for "constraining" blocks. I think this is because a certain axiom emitted by vercors is not applied by viper/z3 (namely, an axiom defining equality between a process and its definition). The (presumably older) version used by vscode seems to have no trouble using the axiom. I think we can either downgrade or figure out a way to tell viper/z3 to use the axiom somehow.

bobismijnnaam:
Fixes #375


wytseoortwijn:
Small note: at this moment version 4.8.6 seems to be more stable with Viper (and VerCors) than 4.8.7.

niomaster:
Can you remove all the headers and bindings? We only need the one executable.

PR #388, opened 2020-02-13
Add Mult case. Fixes #386

### LINKED ISSUE #386: Multiplication with fracs is not handled correctly
Two bugs. First, given the following code:

```
class MyClass {
  int y;

  given frac s;
  requires Perm(x.y, 1\2);
  void foo(MyClass x) {
    frac f;
    frac g = f * 2;
  }
}
```

I expect Vercors to succeed. Instead it throws a lengthy error from Z3:

```[debug] java.lang.AssertionError: assertion failed: Expected both operands to be of the same sort, but found frac (g__2@15@07) and Int (f__1@13@07 * 2).```

This is because the following code is generated:

```
method method_MyClass_foo__MyClass(diz: Ref, current_thread_id: Int, globals: Ref, x: Ref, s: frac)
  requires diz != null
  requires 0 <= current_thread_id
  requires acc(x.field_MyClass_y, 1 / 2)
{
  var f__1: frac
  var g__2: frac
  g__2 := f__1 * 2
}
```

I would expect it to generate this:

```
method method_MyClass_foo__MyClass(diz: Ref, current_thread_id: Int, globals: Ref, x: Ref, s: frac)
  requires diz != null
  requires 0 <= current_thread_id
  requires acc(x.field_MyClass_y, 1 / 2)
{
  var f__1: frac
  var g__2: frac
  g__2 := new_frac(frac_val(f__1) * 2)
}
```

Which works since viper supports multiplication of permissions. To fix this I think `getFracVal` in SilverClassReduction.java has to be made aware that `node` can also be a multiplication. However it might not be the right place to handle this transformation. If someone can confirm that it is, I can implement the fix. (Or, if the cause is something else, someone needs to help me figure out what is)

bobismijnnaam:
Adds a Mul case and ensures fracs are properly packed and unpacked when multiplying. Fixes #386.

### DIFF
--- a/src/main/java/vct/col/rewrite/SilverClassReduction.java
+++ b/src/main/java/vct/col/rewrite/SilverClassReduction.java
@@ -481,6 +481,7 @@ public class SilverClassReduction extends AbstractRewriter {
     case Div:
     case Plus:
     case Minus:
+    case Mult:
     case LT:
     case LTE:
     case EQ:


bobismijnnaam:
One travis job timed out the first time, but succeeded after restarting it. Unsure if this should be merged.

niomaster:
No way to find out unless we try more builds in the wild, so let's just see what happens, that's what dev is for :)

Issue #376, opened 2020-02-01
Splitting read permissions


Drevanoorschot:
I found out VerCors gives an unsufficient permission error where it shouldn't in these two examples:

https://gist.github.com/Drevanoorschot/bc7d211e9f7e4b00f0399dd80b35d4c8
https://gist.github.com/Drevanoorschot/9b15e807a2538d24deee1931877e2632

The idea is to give every thread in the parallel block read permission over the entire input array, which I think should be possible.

I was wondering if this is a bug or if I'm doing it wrong.

niomaster:
In the second example I guess z3 doesn't like the expression 1/(2*t); it does verify when you use (1/2)/t.

wytseoortwijn:
Both examples use non-linear arithmetic (note that handling `read` permissions also involves non-linear arithmetic under the hood). Z3, and thereby also VerCors, aren't particularly good in handling non-linear arithmetic. (See also #377.)

niomaster:
Things we discussed that we might want to do, in order of priority:

* Make a note about non-linear arithmetic in the PVL syntax near operators that make sense
* Make a VerCors tutorial (#88) and include an explanation of this in the tutorial
* Investigate options of Z3 and/or add a flag to use the solver above
* Investigate whether it would be useful to set solver intrinsics per proof goal/method

niomaster:
For now I'll mark the issue as doc, as that's what we assigned most priority.

bobismijnnaam:
For completeness, it seems that the idiomatic vercors way of writing code like this is the following (modulo requires/context_everywhere):

```
class MyClass {
  given frac f;
  requires 0 < f;
  requires 0 < t;
  requires xs != null;
  requires (\forall* int i; i >= 0 && i < xs.length; Perm(xs[i], f));
  void parRead(int[] xs, int t) {
    par parRead(int tid=0..t)
    requires xs != null;
    requires (\forall* int i; i >= 0 && i < xs.length; Perm(xs[i], f\t));
    {
      //some Code
    }
  }

  void foo() {
    int[] xs = new int[10];
    parRead(xs, 3) with { f = 23\47; };
  }
}
```

It doesn't verify currently because a jspec rule is missing to factor out the quantification over tid in the second forall, but once we add that it should work.

PR #389, opened 2020-02-13
Add rewrite rule specifically for dre van oorschot's example in #376

bobismijnnaam:
Complements #376.


niomaster:
you might say it's a

niomaster:
drewrite rule

bobismijnnaam:
@niomaster Would you mind having a long hard look at the rewrite rule before we merge this? Just to make sure this rule doesn't introduce unsoundness.

niomaster:
Looks sound, one note:

* I think you can loosen e6 to a fractional type (ints are subclass of rationals in spec)

PR #394, opened 2020-02-18
Add pure methods & ghost variable support to the C frontend

### LINKED ISSUE #368: Pure functions support for C
It would be very handy to be able to declare pure functions in the specs of a C code, just as in pvl. This should allow to use them in requirements and ensure clauses making verification way more clean.

I do no recall if we discussed about the meaning of pure functions and I do not know if in pvl we actually check that a pure declared function is in fact pure.

I suppose that a good definition of pure follows the 2 basic principles that for same input same output and no side effects (i.e. no modification of outside world).

Anyway I think this is not much needed, and the pure modifier is just something to help you keep the verification clean and you should make sure by yourself that the function is in fact pure if you annotate it so.

### LINKED ISSUE #369: Ghost variable support in C is broken
The following C verification code fails to pass because the ghost variable passed by main into foo is not detected. The error I get is `At file gost_variables.c from line 23 column 5 until line 23 column 10: argument p is missing`

```
/*@
    given frac p;
    context \pointer(a, 1, p);
    ensures \result == 2*(*a);
@*/
int foo(int *a){

    int b;
    b = *a;
    b = b*b;
    return b;
}

//@ given frac q;
//@ context \pointer(c, 1, q);
void main(int *c){

    foo(c) /*@ with {p=q/2;} @*/ ;
}
```

niomaster:
resolves #368, resolves #369

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,29 @@ vct-tool.jar
 
 # Vim swap file
 *.swp
+
+# Haskell/Cabal rules
+dist
+dist-*
+cabal-dev
+*.o
+*.hi
+*.hie
+*.chi
+*.chs.h
+*.dyn_o
+*.dyn_hi
+.hpc
+.hsenv
+.cabal-sandbox/
+cabal.sandbox.config
+*.prof
+*.aux
+*.hp
+*.eventlog
+.stack-work/
+stack.yaml.lock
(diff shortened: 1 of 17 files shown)


raulmonti:
resolves #394

PR #401, opened 2020-02-21
Move language specific passes to parsers

bobismijnnaam:
There are a few passes that are language specific, such as `java_resolve` and `pvl_encode`. Since they are language specific, if possible they should take place in the respective language specific parsers, i.e. `ColJavaParser` and `ColPVLParser`.


bobismijnnaam:
Currently language-specific passes over multiple input programs is not easily possible. Once this is improved this pull request can be reconsidered.

Issue #391, opened 2020-02-14
elect.pvl does not verify: Head operator


petravandenbos:
elect.pvl does not verify. I get the following debug information:

[debug] At Main.java:675: 
[debug] hre.lang.HREError: cannot map operator Head
[debug] At Main.java:675: 
[debug] 	at vct.silver.SilverExpressionMap.map(SilverExpressionMap.java:169)
[debug] At Main.java:675: 
[debug] 	at vct.col.ast.expr.OperatorExpression.$anonfun$accept_simple$2(OperatorExpression.scala:60)
[debug] At Main.java:675: 
[debug] 	at vct.col.util.VisitorHelper.handle_standard(VisitorHelper.scala:22)
[debug] At Main.java:675: 
[debug] 	at vct.col.util.VisitorHelper.handle_standard$(VisitorHelper.scala:22)
[debug] At Main.java:675: 
[debug] 	at vct.col.ast.expr.OperatorExpression.handle_standard(OperatorExpression.scala:10)
[debug] At Main.java:675: 
[debug] 	at vct.col.ast.expr.OperatorExpression.accept_simple(OperatorExpression.scala:60)
[debug] At Main.java:675: 
[debug] 	at vct.col.ast.generic.ASTNode.apply(ASTNode.java:189)
[debug] At Main.java:675: 
[debug] 	at vct.silver.SilverExpressionMap.map(SilverExpressionMap.java:90)
[debug] At Main.java:675: 
[debug] 	at vct.col.ast.expr.OperatorExpression.$anonfun$accept_simple$2(OperatorExpression.scala:60)
[debug] At Main.java:675: 
[debug] 	at vct.col.util.VisitorHelper.handle_standard(VisitorHelper.scala:22)
[debug] At Main.java:675: 
[debug] 	at vct.col.util.VisitorHelper.handle_standard$(VisitorHelper.scala:22)
[debug] At Main.java:675: 
[debug] 	at vct.col.ast.expr.OperatorExpression.handle_standard(OperatorExpression.scala:10)
[debug] At Main.java:675: 
[debug] 	at vct.col.ast.expr.OperatorExpression.accept_simple(OperatorExpression.scala:60)
[debug] At Main.java:675: 
[debug] 	at vct.col.ast.generic.ASTNode.apply(ASTNode.java:189)
[debug] At Main.java:675: 
[debug] 	at vct.silver.SilverExpressionMap.map(SilverExpressionMap.java:494)
[...]

PR #405, opened 2020-02-26
adapted frac_eq and zfrac_eq axioms such that elect.pvl verifies

petravandenbos-utwente:
solves #391


petravandenbos-utwente:
> Can you change the suite of elect? Currently it's in `problem-fail` so it should be in something else if we want it to be in the regular test suite.

Do you want me to just remove the line `//:: suite problem-fail`?

niomaster:
No, it should be in some suite that makes sense I think. Maybe there's one for Future examples?

niomaster:
Never mind, many examples are missing the suite so I think it's fine to remove it :)

PR #415, opened 2020-03-09
Add struct and typedef support

niomaster:


niomaster:
The problem now lies in `StripUnusedExtern`, which is skipped because the methods etc. are no longer wrapped in a class in the C frontend.

niomaster:
Although support hasn't improved in #629, I don't think this is worth keeping around.

PR #425, opened 2020-03-20
Change RangeSeq syntax from [a..b) to {a..b}

niomaster:
Note this PR doesn't include the proposed {a..=b} syntax, because it wans't in vercors anyway.


raulmonti:
Don't you think that () are more usual for ranges than {}? Even though {} are used for sequence notations, I think that what you expect to find inside them are elements of the sequence instead of indexes? Am I wrong? An other option is to have \sum to take two integers instead of a range as first argument.

Issue #431, opened 2020-03-25
ListAppend no longer verifies


Sophietje:
[ListAppend.java](https://github.com/utwente-fmt/vercors/blob/dev/examples/witnesses/ListAppend.java) no longer verifies even though it should pass (according to a comment in the file).

When I try to verify the file with silicon as a backend, it results in the following error:
```
[warning] (class vct.col.util.FeatureScanner): at file examples/witnesses/ListAppend.java from line 19 column 7 until line 115 column 1: detected inheritance
[warning] missing origin
Verification aborted exceptionally: java.util.concurrent.ExecutionException: java.util.NoSuchElementException: None.get
  Cause: java.util.NoSuchElementException: None.get
Errors! (4)
[warning] unspecified error exceptional.error
The final verdict is Fail
```

bobismijnnaam:
This error happens because the resource is not final. What happens now is that the resource is encoded as an APF, which means the "list" predicate becomes an abstract predicate (this is a bug, because the class is marked as final. But apparently something changed since the last time I broke it). Since it is now an abstract predicate, you cannot unfold it, but the example does this (because it expects the predicate to be "normal"). VerCors should give an error in this case, not an exception. When I improve support for APFs this exception should disappear (and turn in a nice error). For now this issue can be fixed by adding "final" to the list resource (since that makes the type of the predicate statically known, which means unfolding works again).

What is more interesting is that this broke without travis complaining. Or did someone merge without waiting for travis, or did I oversee this exception the last time I broke it? (Since I modified inheritance & APF machinery a few weeks ago).

bobismijnnaam:
For reference, this error and its workaround is now documented in the tutorial: https://github.com/utwente-fmt/vercors/wiki/Resources-and-Predicates#vercors-crashes-when-predicates-are-unfolded

PR #432, opened 2020-03-25
Fixed ListAppend example

Sophietje:
examples/witnesses/ListAppend.java no longer verified with the current dev-branch. This PR solves this by adding the "final" keyword to the resource as suggested by @bobismijnnaam . Fixes issue #431

### DIFF
--- a/examples/witnesses/ListAppend.java
+++ b/examples/witnesses/ListAppend.java
@@ -22,7 +22,7 @@ final class List {
   public List next;
   
   /*@
-    public resource list(seq<int> c)=Perm(val,1)**Perm(next,1)**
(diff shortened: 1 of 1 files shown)


raulmonti:
Hi @Sophietje, I am about to merge the pull request. I just wanted to know if this final keyword is being used as a workaround or if it is actually the correct thing to do in this case?

bobismijnnaam:
If I may add my opinion: if proper inheritance support were available in VerCors, the `final` keyword would not be necessary. However, since inheritance support in VerCors is only implemented for about 20%, the `final` keyword is needed. So I would say it is a workaround to turn of inheritance for this example temporarily.

PR #437, opened 2020-04-08
Rework C, Java and PVL parsers

niomaster:
This is a complete rewrite of the parsing infrastructure in VerCors.
## Breaking Changes
- You can no longer split a single specification over multiple line comments, e.g. from ```floats/TestHist.java```, this is now invalid:
```
//@ loop_invariant (\forall int i; 0 <= i && i < hist.length ;
//@                   hist[i]==\sum({0 .. k},\vcmp(vals,\vrep(i)))  );
```
You must instead write:
```
/*@ loop_invariant (\forall int i; 0 <= i && i < hist.length ;
                      hist[i]==\sum({0 .. k},\vcmp(vals,\vrep(i)))  ); @*/
```

- You can no longer write ghost code without ```ghost``` in front, e.g. from ```arrays/JavaArrayExamples.java```, this is now invalid:
```
//@ S1:if(i>0){ recv 0 < i ** i < a.length ** Perm(a[i],1\2) from S2,1; }
```
You must instead write:
```
//@ ghost S1:if(i>0){ recv 0 < i ** i < a.length ** Perm(a[i],1\2) from S2,1; }
```
Mixing statements from the specification grammar and frontend language is still fine. A similar thing goes for ghost variables, which are now declared like ```//@ ghost seq<int> xs;``` instead of ```//@ seq<int> xs;```.

## New Features
- Keywords from specifications are now allowed in the language frontend, and are dealt with correctly;
- You can disable specification lines in comment blocks, with line comments;
- The *ToCOL rewriters should be more readable now, and easier to adapt to future changes;
- The parser now fully deals with parsing specifications in the first pass, so placing specification constructs in new places should be much easier.


raulmonti:
So exiting! :D:D:D:D

niomaster:
All done! Happy to receive any feedback. I have processed all the sonarcloud issues I agree with, but I'm happy to discuss further changes also.

PR #444, opened 2020-05-13
Refactor the toplevel boolean in the Flatten class to use a LevelStack instead.

Drevanoorschot:


niomaster:
I suggested Dre ping you Bob because you looked at this last I think (based on the comment at `toplevel`), but let me know if you don't have time :)

bobismijnnaam:
As a side note, is the flatten phase idempotent now? Not sure if we've discussed this before but might be nice to have. If not, maybe we can open an issue and discuss it there (unless I'm forgetting something and it's very hard/undesirable)

Drevanoorschot:
It is still idempotent as far as I know. The determination of whether something is `toplevel` has changed from altering a flag that determines whether something is `toplevel` to having a stack of `toplevel` determinations per level of the AST. The behaviour is still the same as before but you now don't have to actively think about setting the `toplevel` flag anymore.

bobismijnnaam:
I don't think it was idempotent before... I think I saw a few examples where extra flattens were introduced, but maybe with your fix they are gone (as it missed a few edge cases before obviously). If I run into them again I'll make an issue.

PR #445, opened 2020-05-25
Lots of non-functional refactoring: split ast to COL subproject and more

niomaster:
All the COL ast classes are now in a subproject, plus some generic AST functionality.
Most of the silver logic now belongs to the viper subproject.
HRE lives in the project root again, because we now set up the dependencies in the root build file.
Removed old unused files, while keeping old experiments in a separate package


niomaster:
I have not written any code in this PR, only removed and moved things, so I think we should disregard the sonar judgement.

PR #446, opened 2020-05-26
Structural equality for COL expressions

niomaster:
(depends on #445)
This pull requests converts the last bits of the expressions nodes to scala case classes, for which equality is by default defined as structural equality (as opposed to the normal java default reference equality). This means we can use expressions in expected ways, e.g throw them into a set and have no duplicates; that kind of thing.

The equality of both `MethodInvokation`s and `NameExpression`s is not very well-founded, as the declaration it points to is not taken into account: that's currently a dynamic part of the COL AST. I'd like to make this a static part of the AST at some point, but that's future work.


raulmonti:
@niomaster would you like to look into the conflicts here? I believe that the dev version is the one to keep but not sure

niomaster:
we have encountered a race condition! :)

PR #450, opened 2020-06-03
always deploy commits to dev as pre-release + some ci cleanup

niomaster:


niomaster:
Btw these credentials are linked to my account (which is not a practical problem today: they are encrypted), but from a long-term mantainability standpoint it might be better to exchange those for longer-standing team members' credentials.

bobismijnnaam:
Very nice! Is the version in build.sbt always supposed to contain "snapshot"? Or is the idea that the dev branch has snapshot there, and the master branch contains just has v1.3.0 there, i.e. that the "snapshot" suffix is removed for full releases?

niomaster:
not married to any particular scheme, but I would think we have SNAPSHOT always on the dev branch, and remove it when merging to master.

bobismijnnaam:
Yeah okay, just wanted to be clear on the exact file contents, but this is fine (as merges to master are quite irregular I guess)

Issue #442, opened 2020-05-06
Java logical xor support


bobismijnnaam:
VerCors does not support logical xor for Java. Example program:
```
class C {
  boolean foo() {
    return false ^ true;
  }
}
```
Output:
```
$ vct-dev --silicon temp.java
[warning] (class vct.col.util.FeatureScanner): at file temp.java from line 1 column 1 until line 5 column 1: detected inheritance
At file temp.java from line 3 column 12 until line 3 column 23: Types of left and right-hand side argument are different (Boolean/Boolean).
The final verdict is Error
```
(JLS reference: https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.22.2)

niomaster:
So if I read it correctly, `&`, `|` and `^` are not only bitwise operators, but also boolean operators when the lhs&rhs are both boolean. This means there is no low-precedence operator, because there is no logical-only xor operator (such as `^^`)

PR #465, opened 2020-06-13
Add Ambigious operators for languages that use the same operators for different data types

Drevanoorschot:
Fixes [issue 442](https://github.com/utwente-fmt/vercors/issues/442). For example, AmbigiousXor can be used to tokenize a `^` in Java to later determine whether it should be a bitwise xor or logical xor.

### DIFF
--- a/col/src/main/java/vct/col/ast/expr/StandardOperator.java
+++ b/col/src/main/java/vct/col/ast/expr/StandardOperator.java
@@ -31,7 +31,13 @@ public enum StandardOperator {
   /** Bitwise eXclusive OR. */
   BitXor(2),
   /** Bitwise negation or complement. */
-  BitNot(2),
+  BitNot(2), //TODO change arity to 1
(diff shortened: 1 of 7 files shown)


bobismijnnaam:
Nitpicky idea: should we rename And, Or, Xor, to something like BoolAnd, BoolOr, BoolXor as well? That way the semantics are explicit.

niomaster:
I would suggest keeping them as is for this pull request, and consider giving them a more explicit name after we make a distinction between regular computational `And` and the conditional `&&`/`||` operator from java, where the right side is only evaluated (and methods invoked etc.) when its resulting value matters. e.g. in `a && f()`, `f` is only invoked when `a` is true.

niomaster:
For posterity: the way it currently works is that for either type of and we emit `&&` to viper. In Viper, `&&` is short-circuiting, but my guess is that we flatten expression to a sufficient degree that we usually don't "use" the short-circuiting. In any case, it's not very well-defined currently.

PR #472, opened 2020-06-19
Set subset operator in PVL

OmerSakar:
This PR adds support for the subset operator in PVL.


examples/sets/subset.pvl
```diff
@@ -0,0 +1,18 @@
...
+
+class SubSet {
+	void main() {
+		set<int> a = set<int> {1, 5, 7, 8, 6, 1, 4, 8, 6, 3};
+		set<int> b = set<int> {1, 5, 7};
+
+		assert b <= b && b < a;
+		
+		bag<int> c = bag<int> {1,5,7,4,9,6,3,2,4,5};
+        bag<int> d = bag<int> {4,6,2,4,5};
```
raulmonti, line 14: indent here

examples/sets/subset.pvl
```diff
@@ -0,0 +1,18 @@
...
+
+        assert d < c;
```
raulmonti, line 16: indent here too

examples/sets/subset.pvl
```diff
@@ -0,0 +1,18 @@
+// -*- tab-width:2 ; indent-tabs-mode:nil -*-
```
raulmonti, line 1: Can we add asserts for the not contains cases? such as asssert !({a,b,c} < {a,b,d,e,f})

viper/src/main/java/viper/api/VerCorsExpressionFactory.java
```diff
@@ -76,6 +76,16 @@ public ASTNode any_set_intersection(Origin o, ASTNode e1, ASTNode e2) {
     return res;
   }
 
+  @Override
+  public ASTNode any_set_subset(Origin o, ASTNode e1, ASTNode e2) {
+      enter(o);
+      //TODO check what to put here.
```
raulmonti, line 82: I believe this is OK since at least Java and C will not have a sets or bags as basic types? thus they do not have this operation? or am I wrong. If not, then we may remove the comment.

examples/sets/subset.pvl
OmerSakar: I guess I could not choose between tabs and spaces. Will be fixed in next commit.

examples/sets/subset.pvl
OmerSakar: I guess I could not choose between tabs and spaces. Will be fixed in next commit.

examples/sets/subset.pvl
OmerSakar: Yes, I will add one for sets and one for bags (in the next commit).

viper/src/main/java/viper/api/VerCorsExpressionFactory.java
OmerSakar: Good that you found the TODO, that should have been removed as the comment is not applicable anymore. Will be removed in the next commit.

PR #477, opened 2020-06-29
Remove useless throws declarations

bobismijnnaam:
Remove throws declarations previously added to test if sonarcloud could detect methods that always throw.


bobismijnnaam:
Closed as it is merged

PR #487, opened 2020-07-15
Add coverage

bobismijnnaam:
This pull request adds support for sonarcloud coverage reporting through integration of Jacoco in the vercors test suite.

To do:
- [x] Change the use of ':' into Java builtin `pathSep`
- [x] Zip class files before uploading, unzip after downloading
- [x] Add windows job for the "basic" folder
- [x] Add mac job for the "basic" folder
- [x] Increase number of workers per job to 2


bobismijnnaam:
I think this PR will be subsumed by #578 and the recent move to github actions (which enables trivial file sharing, which was not possible before). Until then we keep this PR around in case it proves useful in the future, but I think it won't be needed if #578 works out.

.github/workflows/build-test-slow-medium.yml.disabled
```diff
@@ -0,0 +1,42 @@
+name: Vercors build and test workflow for the slow and medium test suites. Currently disabled as some work is needed to get the slow/medium suites to pass again.
```
Vescatur, line 1: Could you remove this document?

.github/workflows/build-test.yml
```diff
@@ -113,8 +119,103 @@ jobs:
...
+      # --strip-components is needed because the zip contains a directory, which contains bin, examples, etc.
+      # But the name of this directory changes from time to time, because of the version number
+      # By stripping the first component of the path when extracting this directory this is not a problem
+      # --directory extracts to a given directory, ensuring the current directory is not cluttered.
+      - run: tar --strip-components=1 --directory vercorszip -xf vercors.zip
+      # Note how arguments with commas are quoted triply:
+      # - One layer gets peeled of by the .bat script
+      # - Another layer gets peeled of by powershell
+      # This prevents the commas from being interpreted as array separators, a marvelous powershell feature.
+      - run: vercorszip\bin\vercors.bat --test=examples\basic --test-workers=2 --tool carbon --tool silicon """--exclude-suite=slow,medium,problem-fail,skip-travis""" --progress --actions-test-output
```
Vescatur, line 168: skip-travis is not needed anymore. 

Could you merge dev into this branch to check whether my changes are compatible with yours?

.github/workflows/build-test.yml
```diff
@@ -113,8 +119,103 @@ jobs:
...
+
+  integration-test-mac:
```
Vescatur, line 170: I think it is nice that we are testing on all the operating systems.

examples/verifythis/2018/challenge1.pvl
```diff
@@ -1,5 +1,3 @@
-// -*- tab-width:2 ; indent-tabs-mode:nil -*-
```
Vescatur, line 1: I will check whether this document is in the correct directory after merging.

examples/verifythis/2018/challenge2.pvl
```diff
@@ -2,8 +2,10 @@
```
Vescatur, line 2: Should be in known-problems which it will probably will be after merging.

Vescatur (changes requested):
There are two jars added to the git repository (jacocoagant.jar and jacococli.jar) . Is that necessary or can they added as a dependency? If you choose to remove them, could you also remove them from the git history to save on storage?

Vescatur:
I think it is easier to merge pull request that focus on one problem. This pull request fixes an example, expanded the test suite so that it runs on other operating systems and does code coverage. I think these separate features could have already been merged with dev before everything was finished.

Vescatur:
> There are two jars added to the git repository (jacocoagant.jar and jacococli.jar) . Is that necessary or can they added as a dependency? If you choose to remove them, could you also remove them from the git history to save on storage?

I see that the project is already 700 mb so might not be really relevant.

bobismijnnaam:
> I think it is easier to merge pull request that focus on one problem. This pull request fixes an example, expanded the test suite so that it runs on other operating systems and does code coverage. I think these separate features could have already been merged with dev before everything was finished.

I agree, this PR got out of hand. I'll pay better attention next time :)

> I see that the project is already 700 mb so might not be really relevant.

All the little bits add up, so I agree that we should be on the lookout for them. But in this case it cannot be avoided I think.

Vescatur (approved):
I appreciate the effort into decreasing the nesting of CommandLineTesting. The rest of the pull request also looks good.

Issue #492, opened 2020-07-17
Cannot map constant value class LongValue


Sophietje:
The following program results in `HREError: cannot map constant value class vct.col.ast.expr.constant.LongValue`:
```java
public class SwapLong {
  long F;
  long G;
  /*@ requires Perm(F,1) ** Perm(G,1);
      ensures Perm(F,1) ** Perm(G,1);
      ensures F == \old(G) && G == \old(F);
   */
  void n()
  {
    long tmp;
    tmp = F;
    F = G;
    G = tmp;
  }
  
  /*@ requires Perm(F,1) ** Perm(G,1);
      ensures Perm(F,1) ** Perm(G,1);
      ensures F == \old(F) && G == \old(G);
   */
  void wrong()
  {
    n();
  }
}
```

Perhaps a missing `LongValue` case in [map(ConstantExpression e)](https://github.com/utwente-fmt/vercors/blob/a54f323e5d360e98aa59f8aeb95e7eba8adba739/viper/src/main/java/viper/api/SilverExpressionMap.java#L61)?

niomaster:
I think rather than pretending we have some distinction between longs and ints we can better just translate long values to IntegerValue's (so unbounded integers) in the parser

bobismijnnaam:
I agree. Maybe we could also emit a warning message whenever a different type than int is used, i.e. "<not int> supported, but no bounds checking for <not int>". Then at least people know we're not doing overflow checking and such once they start types different from ints.

niomaster:
They are now all translated to the same type and constant value node.

Issue #428, opened 2020-03-24
Find out which backends are still supported


Sophietje:
Currently we are always using the silicon backend to verify our programs. However, we used to also support backends such as Dafny, Chalice and Boogie. We need to figure out if these are still supported or not. To check this, we can try out the examples in the examples directory that are annotated with "//:: tools X" (where X is not silicon).

After this has been done, if some backends turn out to be no longer supported, we can decide whether we'd like to add this support again (if so, make a new issue). Otherwise, we'll need to try and salvage the examples by converting them to something that silicon can verify.

Sophietje:
Update:
We have examples that use the backend dafny, carbon, boogie and chalice (or silicon). All examples that are annotated with "//:: tools dafny", "//:: tools carbon", "//:: tools boogie", "//:: tools chalice" currently fail. 

There was also a file that mentioned "it works with verifast", not sure if that means that we also supported verifast as a backend?

Sophietje:
As promised, here is a list of all examples that used either dafny, carbon, boogie or chalice as a backend. Trying to verify one of these examples with the specified backends will result in an error. Below you can find the list of failing examples, they are grouped by which error occurs when trying to verify it:

**Syntax error (no viable alternative at input X), HREExitException**

- [ ] examples/inheritance/FullThreadInstance.java examples/inheritance/FullThread.java examples/inheritance/FullThreadMain.java (--chalice --explicit)
- [ ] examples/atomics/DWLock.java (--chalice --explicit)
- [ ] examples/inheritance/Fibonacci.java examples/inheritance/FullThread.java (--chalice --explicit)
- [ ] examples/witnesses/ListIterator.java (--chalice)
- [ ] examples/witnesses/WitnessDemo.java (--chalice)
- [ ] examples/witnesses/predicates.pvl (--chalice)
- [ ] examples/lists/linkedlist.pvl (--chalice, this one does not result in a HREExitException)


**Constructor has a different name than the class in which it is defined**

- [ ] examples/atomics/RBLock.java (--chalice --explicit)


**Unresolved name X found during type check**

- [ ] examples/manual/witness.pvl (--chalice)


**Could not find method X (in some of these the 'method' it cannot find is actually a predicate)**

- [ ] examples/witnesses/Twice.java (--chalice --explicit)
- [ ] examples/inheritance/SimpleThreadInstance.java examples/inheritance/SimpleThread.java examples/inheritance/SimpleThreadMain.java (--chalice)
[...]

PR #493, opened 2020-07-20
Fixes several Java examples

Sophietje:
I skipped examples that use witnesses since we still have to figure out support for those. 

Note that there are some examples in here that do not yet verify due to missing support for resource@Class (APFs/inheritance), as well as an issue with longs (see #492). I have fixed the syntax of these examples so they can be used for testing when the issues/missing support are resolved. 

Some of these examples used boogie/chalice/dafny as a backend. I have converted those that I could such that they can be verified with the silicon backend. As a result, this PR fixes some of the examples mentioned in #428.


Sophietje:
Travis failed on the LeaderElect example which I didn't touch in this PR...

Issue #217, opened 2019-05-16
Pure function calls inside invariants


wytseoortwijn:
The following code snippet exhibits a bug in VerCors:

```java
class Foo {
  invariant xs != null;
  requires (\forall* int i; 0 <= i && i < xs.length; Perm(xs[i], 1/2));
  ensures \result;
  pure boolean f(int[] xs);

  invariant xs != null;
  requires (\forall* int i; 0 <= i && i < xs.length; Perm(xs[i], 1));
  void bar(int[] xs) {
    invariant inv((\forall* int i; 0 <= i && i < xs.length; Perm(xs[i], 1/2)) ** f(xs))
    {
      par (int tid = 0 .. xs.length)
        context Perm(xs[tid], 1/2);
      {
        atomic(inv) { } // problem is here
      }
    }
  }
}
```

Inside the `invariant` is a function call `f(xs)`, with `xs` an integer array. The function `f` requires read permission for each entry in `xs`. This causes VerCors to crash, due to a problem in `atomic(inv) { }`. More specifically, VerCors gives:

```
class Foo
  method f is final
    initial declaration
  method bar is final
    initial declaration
  method Foo is final
    initial declaration
=== ./examples/test/test25.pvl     ===
  11 		
  12 		invariant inv(
                                                                      [-----
  13 			(\forall* int i; 0 <= i && i < xs.length; Perm(xs[i], 1/2)) ** f(xs)
                                                                       -----]
  14 		)
  15 		{
-----------------------------------------
  Local name globals is undefined
=========================================
At file ./examples/test/test25.pvl from line 13 column 67 until line 13 column 71: fatal error
Triggered by file ./examples/test/test25.pvl from line 13 column 67 until line 13 column 71:
The final verdict is Error
entire run took 990 ms
```

This has something to do with permissions I think. If `f` does not require any permissions, everything goes well.

sjcjoosten:
Related to #194. Did #201 fix this? (Can someone test this again and close if it doesn't reproduce)

Sophietje:
I pulled the newest version of VerCors and ran it on the example above. It still gives me an error, though it is a different error this time.

```
=== ../../VerCors-Examples/20190718-testIssue217.pvl ===
   8   requires (\forall* int i; 0 <= i && i < xs.length; Perm(xs[i], 1));
   9   void bar(int[] xs) {
                                                                                     [-----
[...]

PR #497, opened 2020-07-27
Refactor NameScanner

bobismijnnaam:
NameScanner scans the AST and collects free variables. I need to know which free variables is only read from and which is written to for issue #217, but NameScanner originally was hard to extend with this functionality. This PR refactors NameScanner to be more flexible, and implements the read/write functionality as well. It also cleans up some code that uses NameScanner.


col/src/main/java/vct/col/ast/util/NameScanner.java
```diff
@@ -1,210 +1,210 @@
...
-
-  @Override
-  public void visit(MethodInvokation invokation) {
-    super.visit(invokation);
-    // The before/after annotations of methods contain the given/yields mappings, so don't include them as though they
-    // are free names.
-    auto_before_after = false;
-  }
-}
+//package vct.col.ast.util;
```
niomaster, line 1: This file can now be removed I think

col/src/main/java/vct/col/ast/util/NameScanner.scala
```diff
@@ -0,0 +1,210 @@
...
+        case Some(value) =>
+          // If the name is already a free variable, the types must match
+          if (t != null && !(value.typ == t)) {
+            Fail("type mismatch %s != %s", t, value.typ);
+          }
+        case None if !hasDecl(name) =>
+          // If it is new and there is no decl, declare a new free variable
+          put(name, t)
+        case _ =>
+          // Otherwise ignore it
```
niomaster, line 109: I don't understand what is happening here, but if this was the old situation let's just leave it.

col/src/main/java/vct/col/ast/util/NameScanner.scala
```diff
@@ -0,0 +1,210 @@
...
+
+  override def visit(s: LoopStatement): Unit = {
+    push()
+
+    super.visit(s)
+
+    pop()
+  }
+
+  override def visit(e: BindingExpression): Unit = if (e.getDeclCount == 1) {
```
niomaster, line 151: Why this limitation?

col/src/main/java/vct/col/ast/util/NameScanner.scala
```diff
@@ -0,0 +1,210 @@
...
+  else Abort("missing case in free variable detection")
+
+  override def visit(s: ForEachLoop): Unit = {
+    push()
+
+    super.visit(s)
+    if (s.get_before != null) s.get_before.accept(this)
+    if (s.get_after != null) s.get_after.accept(this)
+    // Ensure before/after are not scanned after pop()
+    auto_before_after = false
```
niomaster, line 167: This seems kinda leaky

col/src/main/java/vct/col/ast/util/NameScanner.scala
```diff
@@ -0,0 +1,210 @@
...
+import vct.col.ast.`type`.Type
+import NameExpressionKind._
+import vct.col.ast.generic.ASTNode
+import vct.col.ast.stmt.composite.VectorBlock
+import vct.col.ast.stmt.terminal.AssignmentStatement
+
+object NameScanner {
+  def freeVars[R <: ASTNode](nodes: util.List[R]): util.Map[String, Type] = {
+    val scanner = new NameScanner
+    for (n <- nodes.asScala) {
```
niomaster, line 23: ```nodes.asScala.foreach(_.accept(scanner))```

col/src/main/java/vct/col/ast/util/NameScanner.scala
```diff
@@ -0,0 +1,210 @@
...
+
+  @varargs
+  def freeVars(theNodes: ASTNode*): util.Map[String, Type] = freeVars(theNodes.asJava)
+}
+
+class NameScanner extends RecursiveVisitor[AnyRef](null, null) {
+
+  case class Entry(name: String, typ: Type, var writtenTo: Boolean)
+
+  val frameStack: mutable.Stack[mutable.Set[DeclarationStatement]] = mutable.Stack()
```
niomaster, line 37: Scala docs: "@deprecated (Since version 2.12.0) Stack is an inelegant and potentially poorly-performing wrapper around List. Use a List assigned to a var instead."

bobismijnnaam: In 2.13.0 it is not deprecated anymore, and the interface fits, so I would like to keep using it. Couldn't find a "@ignore-deprecated-warning" annotation or something similar.

col/src/main/java/vct/col/ast/util/NameScanner.scala
bobismijnnaam: I've rewritten the match as there was also an edge case missing and added a comment, hopefully it is clearer now

col/src/main/java/vct/col/ast/util/NameScanner.scala
bobismijnnaam: It was also in the original code, so that's why I left it in. Not sure if it has ever been a problem? Need to investigate a bit if it matters if we want to take it out

col/src/main/java/vct/col/ast/util/NameScanner.scala
bobismijnnaam: Yeah, and I'm not sure why ForEachLoop does get to scan before/after, and not ParallelBlock and MethodInvokation. (the auto_before_after is meant to be used this way, so that part is okay I think as long as the flag itself exists)

col/src/main/java/vct/col/ast/util/NameScanner.scala
bobismijnnaam: Could not think/find any signs that this is still needed. I've changed it, will investigate if travis complains.

col/src/main/java/vct/col/ast/util/NameScanner.scala
bobismijnnaam: Even more so: I'd argue that with/then blocks _should_ be scanned, because if they use free variables those should be taken into account as well. Only the target variable names within given/yields should be ignored.

bobismijnnaam: The reason why it is this way (I think) is that if there is a decl in the before block, it should only be in scope for the for loop and the after block. Re: par blocks, they don't have before/after, so I removed it there. Re: method invokations and with/then, there I will additionally implement that they are incorporated as well, taking into account that given/yields variables should not be recorded as free variables.

col/src/main/java/vct/col/ast/util/NameScanner.scala
```diff
@@ -60,15 +86,39 @@ class NameScanner extends RecursiveVisitor[AnyRef](null, null) {
...
+    * If there is a free variable with the same name, the types must match
+    * If there is no decl nor a free variable, we found a new free variable, which is added to the `freeNames` list
+    *
+    * @param additionalNames can be used to pass in given or yields scopes.
+    * @return true if `name` is the name of a free variable.
+    */
+  private def checkName(name: String, typ: Type, additionalNames: Function1[String, Option[DeclarationStatement]]): Boolean =
+    (additionalNames(name) orElse getDecl(name), freeNames.get(name)) match {
+      case (Some(decl), _) =>
+        if (typ != null && decl.`type` != typ) {
```
niomaster, line 99: You may prefer `Option(typ)`, which is equivalent to `match typ { null => None; other => Some(other) }`

niomaster: and then `Option(typ).isEmpty` or whatever, but it may just be more verbose, idk.

col/src/main/java/vct/col/ast/util/NameScanner.scala
```diff
@@ -36,14 +34,42 @@ class NameScanner extends RecursiveVisitor[AnyRef](null, null) {
...
-  push()
+  pushFrame()
 
-  private def hasDecl(name: String): Boolean = frameStack.exists(frame => frame.exists(p => p.name == name))
+  /**
+    * Finds the first declaration in the framestack that has name `name`
+    */
+  private def getDecl(name: String): Option[DeclarationStatement] = frameStack
+    // Unlift turns a function returning Option into a partial function
+    .collectFirst(Function.unlift(_.find(decl => decl.name == name)))
```
niomaster, line 57: While this may be very nice functionally, I don't think it's very readable (I still had to look up what exactly a partial function in scala is etc.). Maybe you could consider:

```
frameStack.collectFirst(frame => frame.find(_.name == name) match {
    case Some(decl) => decl
})
```

I think that still produces a partial function with respect to the match, or maybe even:

```
frameStack.map(_.find(_.name == name)).collectFirst { case Some(decl) => decl }
```

niomaster: I could not resist making an unreadable functional implementation:

```
frameStack.map(_.find(_.name == name)).reduce(_ orElse _)
```

col/src/main/java/vct/col/ast/util/NameScanner.scala
```diff
@@ -60,15 +86,39 @@ class NameScanner extends RecursiveVisitor[AnyRef](null, null) {
...
   }
 
+  /** If the name is in `additionalNames`, or there is a regular declaration with the same name, the types must match
+    * If there is a free variable with the same name, the types must match
+    * If there is no decl nor a free variable, we found a new free variable, which is added to the `freeNames` list
+    *
+    * @param additionalNames can be used to pass in given or yields scopes.
+    * @return true if `name` is the name of a free variable.
+    */
+  private def checkName(name: String, typ: Type, additionalNames: Function1[String, Option[DeclarationStatement]]): Boolean =
```
niomaster, line 96: `Function1` can just be `String => Option[DeclarationStatement]` I think? Or is there a subtle difference?

niomaster: And shouldn't it be a map? I guess that's applicable also haha

col/src/main/java/vct/col/ast/util/NameScanner.scala
niomaster: or even `frameStack.map(_.find(_.name == name)).flatten.headOption`, so many options!

niomaster: more seriously, I think scala docs imply that the order of iterating over a stack is FIFO, I don't think that is what you want?

bobismijnnaam: The main idea for using unlift is to prevent looking in every frame of the stack. However I agree it's a bit cryptic, it's probably better to change it to `frameStack.map` and eat the performance penalty. (I think your very first example is the most clear, but since it starts with `frame => ...` it is still a regular (total?) function and not a partial function, so it wouldn't work I think. But I'll doublecheck it)

The iteration starts at the top of the stack, which means the most recent declaration is found first if I understand correctly. I think that is what we want here.

col/src/main/java/vct/col/ast/util/NameScanner.scala
bobismijnnaam: I am not sure how Option(typ) improves it, I guess it can replace `typ != null`? But I think that doesn't make it much clearer, so if its okay with you I will keep it as is.

col/src/main/java/vct/col/ast/util/NameScanner.scala
bobismijnnaam: I didn't know how to write a function type, `string => option` is what I intended. I guess ideally it would be a map, but since we need to take nesting into account stacks of maps are used. Therefore a function is better here for flexibility. (Or does scala understand that Map[String, Option[Decl]] can also be a function? When I tried it it just gave errors at the call sites)

PR #501, opened 2020-07-29
Allow pure function calls in invariants and check local variable usage

### LINKED ISSUE #217: Pure function calls inside invariants

### LINKED ISSUE #418: Generated viper code contains <.*>
When generating viper code, the method names contain stuff like "<Array<Cell<Edge>>>". This causes the viper plugin to complain about < and > (parse error). Therefore, it'd be great if we could remove < and > from the generated code.

bobismijnnaam:
Fixes #217:

1. Moves the atomic inlining pass to before parallel_block_encoder so it properly detects free variables instead of introducing new free variables during parallel_block_encoder
2. Uses the new namescanner to ensure parallel blocks do not write to variables from which other parallel blocks might read
3. Local variables that occur in invariants can not be written to within the invariants

2 and 3 in combination also ensure that a parallel block does not invalidate an invariant established in another parallel block.

This PR also fixes #418 by replacing \< and \> with $ in `silver-class-reduction`.


raulmonti (approved):

raulmonti:
@bobismijnnaam there is a single file conflict here. @niomaster if you have some time please also take a look at the PR so we can be more confident on its correctness.

PR #512, opened 2020-08-16
Rework main

niomaster:


raulmonti:
@niomaster there is a conflict here, can't merge :S

niomaster:
I've continued more significant work on this in a different branch that now also conflicts, so I'd prefer to resolve the merge conflicts there and merge that instead, later.

PR #516, opened 2020-08-25
Remove identity operator from the standard operators

### LINKED ISSUE #239: Generated identity function
VerCors seems to generate/add an identity function to every Viper file it produces (at least in my older, late 2018 version, it does):
```java
function id(x: Int): Int 
{
  x
}
```
However, I notices that this function sometimes makes verification harder. For example, in my version of VerCors, the following code snippet does not verify, and this is only because the `id` function is being used:
```java
invariant a != null;
invariant a.length % 2 == 0;
context (\forall* int j; 0 <= j && j < a.length; Perm(a[j], write));
void zeroall(int[] a) {
  par (int tid = 0 .. a.length/2)
    context Perm(a[2 * tid], write);
    ensures a[2 * tid] == 0;
  {
    a[2 * tid] = 0;             
  }
}
```

Could we get rid of `id`, or make it into a macro, instead of a function, so that Viper can easily eliminate it?

Drevanoorschot:
Removes identity operator as suggested in issue #239. Also closes #239.


parsers/src/main/java/vct/parsers/CMLtoCOL.scala
```diff
@@ -1061,8 +1061,6 @@ class CMLtoCOL(fileName: String, tokens: CommonTokenStream, parser: CParser)
       create expression(Length, expr(exp))
     case ValPrimary14("\\old", "(", exp, ")") =>
       create expression(Old, expr(exp))
-    case ValPrimary15("\\id", "(", exp, ")") =>
-      create expression(Identity, expr(exp))
```
niomaster, line 1065: Can you also remove it from the specification grammar?

niomaster (changes requested):

parsers/src/main/java/vct/parsers/CMLtoCOL.scala
Drevanoorschot: @raulmonti I removed it from the specification grammar as requested. It should be ok to merge now (I think).

niomaster (approved):
All good, I expect the ValPrimary numbering to have changed now, but travis will probably indicate that if so.

parsers/src/main/java/vct/parsers/CMLtoCOL.scala
raulmonti: Something broke :S no tests are passing @Drevanoorschot  and @niomaster

Drevanoorschot:
it was indeed the case that the numbering changed.I fixed the numbering in the parsers so it should at least be able to build now

niomaster:
Just so you know by the way, we should be using this feature from antlr: https://github.com/antlr/antlr4/blob/master/doc/parser-rules.md#alternative-labels

Instead of the numbered rules, we would get the alternative name that we choose so we don't need to shuffle things around every time.

raulmonti:
Looks like it still crashes? @niomaster what was the difference between the two travis tests? is one running on the target branch the other on the the expected result of the merging?

bobismijnnaam:
> Looks like it still crashes? @niomaster what was the difference between the two travis tests? is one running on the target branch the other on the the expected result of the merging?

That is correct.

niomaster:
This is just a situation where git things a merge can be automatic, when it in reality can't be automatic. The proper solution is therefore to merge dev into this branch, fixing the problem in the merge commit.

bobismijnnaam:
I guess the proper command in this case would be `git merge dev --no-commit` (so you can tweak the merge results before the commit is made)? Or is there a different strategy (besides rebasing)?

niomaster:
Yes, --no-commit would be the way to go.

bobismijnnaam:
I never heard any discussion about this, so maybe I should just ask: branches should be deleted after merging the related PR, unless they're needed for further work, right? Or is there a reason to keep them around? (I used to forget about my branches often, but seeing as we have quite a lot of branches now I usually delete them afterwards) Possibly we can document this policy somewhere but that doesn't help as much as just reminding each other regularly I think.

niomaster:
Yes (I just delete them every now and then for everyone)

PR #517, opened 2020-08-26
Add monotonic bool example that uses the new forall read squash rule

bobismijnnaam:


examples/parallel/monotonicBool.pvl
```diff
@@ -0,0 +1,82 @@
+//:: case MonotonicBool
+//:: tool silicon
+
```
Sophietje, line 3: Maybe also add the 'verdict' and 'suite' part of the header?

PR #526, opened 2020-09-02
Viper 20.07

niomaster:


examples/openmp/addvec1.pvl
```diff
@@ -1,6 +1,8 @@
 // -*- tab-width:2 ; indent-tabs-mode:nil -*-
 //:: case ParallelVectorAdd1
 //:: tools silicon
+// fails due to https://github.com/viperproject/silicon/issues/512
+//:: suite problem-fail
```
bobismijnnaam, line 5: Just for my information: what is the intention of the problem-fail suite? We keep tests there that are problematic but that we want to fix at some point, and we check them whenever viper updates if they start working again?

bobismijnnaam (approved):
Nice!!!!

PR #528, opened 2020-09-04
Predicate refactor

bobismijnnaam:
Improve support for predicates.

- Inline predicates can be scaled.
- Add tests for inline predicates.
- Unfolding and folding of inline predicates is now encoded as asserting the predicate body.


src/main/java/vct/col/rewrite/InlinePredicatesRewriter.java
```diff
@@ -68,43 +84,97 @@ public void visit(Method m){
...
         break;
     }
   }
+
+  @Override
+  public void visit(ASTSpecial e) {
+      switch (e.kind) {
+        case Fold:
+        case Unfold:
+          if (inline(e.getArg(0))) {
```
niomaster, line 169: Still think we should discuss this in the next meeting :)

niomaster (approved):

src/main/java/vct/col/rewrite/InlinePredicatesRewriter.java
bobismijnnaam: Yes it is in my notes! ;)

niomaster:
The decision from the meeting was to consider unfolding or folding an inline predicate a no-op, but warn that the statement is meaningless.

raulmonti:
@bobismijnnaam could you look into this PR and update it?

bobismijnnaam:
Continued at #639.

Issue #467, opened 2020-06-16
Flattening ternary operator breaks short-circuit


ArmborstL:
In the following example
```java
/*@
  requires x > 0;
@*/
int log2(int x) { 
  return x/2 > 0 ? log2(x/2)+1 : 0; 
}
```
the return statement gets flattened by VerCors into the following Viper expression:
```java
__flatten_5__6 := method_T_log2__Integer(diz, current_thread_id, globals, x / 2)
sys__result := (0 < x / 2 ? __flatten_5__6 + 1 : 0)
```
In Java (and many other languages, including Viper), the ternary operator is short-circuit, i.e. only those parts of the expression are evaluated which are needed. In particular, the recursive call is only performed if the condition holds.
The flattening breaks this, performing the recursive call unconditionally. Thereby, the precondition of the recursive call can no longer be guaranteed.
To avoid this problem, the flattening should be conditional, i.e. using an `if`.

PR #529, opened 2020-09-07
Add if-then-else flatten case that converts it to an if statement

Drevanoorschot:
This pull request fixes issue #467.


raulmonti:
SonarCloud complains because we are updating a static field from a non-static non synchronised method in 

        String name = "__flatten_"+(++counter);

Do you think we should fix this or should we just ignore it since it doesn't look that dangerous?

niomaster:
We should ignore this; of course it's not good we need global state, but this prevents e.g. overlapping names if we apply flatten twice.

PR #531, opened 2020-09-11
logging may overwrite progress messages; make test progress list currently running tasks

niomaster:
Progress lines were printed as:

```
"[progress] %s%n"
```

But now are printed as:

```
"[progress] %s\r"
```

Any lines following a progress line will start with "\033[0K", a widely supported [ansi escape code](http://ascii-table.com/ansi-escape-sequences.php) to clear the current line. The effect is that progress lines are visible only until the next line is printed, usually the next progress line.

### DIFF
--- a/hre/src/main/java/hre/lang/System.java
+++ b/hre/src/main/java/hre/lang/System.java
@@ -147,6 +147,8 @@ public class System {
         debugfilterByLine.add(classLineCombo);
     }
 
+    private static boolean needLineClear = false;
+
     private static void log(LogLevel level, Map<Appendable, LogLevel> outputs, String format, Object... args) {
         // Only format the string (expensive) when the message is actually outputted
         String message = null;
@@ -158,11 +160,22 @@ public class System {
                         for (LogWriter writer : activeLogWriters) {
                             writer.doFlush();
                         }
+                        message = level.getShorthand() + String.format(format, args);
+                    }
 
-                        message = level.getShorthand() + String.format(format + "%n", args);
+                    if(needLineClear) {
+                        entry.getKey().append("\033[0K");
+                        needLineClear = false;
                     }
 
                     entry.getKey().append(message);
+
+                    if(level == LogLevel.Progress) {
+                        entry.getKey().append("\r");
+                        needLineClear = true;
+                    } else {
+                        entry.getKey().append("\r\n");
+                    }
                 } catch (IOException e) {
                     if (level != LogLevel.Abort) {
                         Abort("IO Error: %s", e);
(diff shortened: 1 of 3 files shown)


niomaster:
Unfortunately it looks like travis doesn't show the line until a line feed is encountered so we might have to disable it there.

niomaster:
@ArmborstL could you check what this looks like in the windows command line? Or do you usually use mingw/git-bash or so?

bobismijnnaam:
> so we might have to disable it there.

Does this imply it can be disabled (e.g. with a flag or something)? If not, I'd like to request a flag to turn it off (since I've used the log of applied passes for debugging on occasion).

ArmborstL:
> 
> 
> @ArmborstL could you check what this looks like in the windows command line?

Not good 😞 
On PowerShell (the Windows-internal shell I usually use), it works in the sense that it overwrites previous progress messages, instead of printing them all below each other. However, it tries to print the special character at the beginning of the line, and if the previous progress line was longer than the current, you still see the last few characters of the previous message after the current one. 
delete-progress_PS

On the old `cmd` command prompt, it is nearly the same, except for an additional progress message at the end. I think the cmd automatically adds an empty line after the output before prompting the next command, while PowerShell does not. That empty line then still has an old progress message.
delete-progress_CMD

niomaster:
Okay, we should not spend a lot of time on pretty logging :laughing:. I'll put it behind a default-disabled flag because I'd personally like to keep it. Thanks for helping out @ArmborstL @bobismijnnaam

ArmborstL:
I looked around a bit more. In principle, cmd and PowerShell support ANSI sequences, and e.g. if I use them in a `txt` file and print the file to the screen, they are parsed successfully. However, when I used them in an `echo` in a batch script, they were not executed (they were skipped, i.e. output was as if they were simply not there). And from within Java, they are printed as characters (see pictures in previous comment). 
Googling a bit, it seems that the support for these sequences is disabled by default. I guess the `cat` command to print a file enables it automatically. To enable it for a custom program, most sources point to either changing the global default in the Windows registry (which I don't think we want to do), or calling an API in C; in Java or other languages it is very cumbersome. 
So I guess @niomaster is right, it's not worth the effort for some pretty printing. But a flag to enable it could be nice, yes 👍

bobismijnnaam:
As a final workaround it could be disabled on windows by default and enabled on linux/mac, as I like the idea of having it on by default too. But that does create some dissonance between the platforms so not sure if it's wise.

niomaster:
These changes will be carried along with #541

bobismijnnaam:
Can this branch be deleted?

PR #534, opened 2020-09-16
Add --notify flag

bobismijnnaam:
For serious case studies verification can take upwards of 10 minutes. This PR adds the --notify flag that makes waiting a bit more bearable by sending a notification when vercors finishes. Currently supported OSs are mac, ubuntu/debian, and windows 10. No new dependencies are introduced so the added functionality remains lightweight (and should remain so, IMHO - this functionality is not worth including heavyweight OS libraries).


niomaster:
Are you still planning on changing the mac support?

bobismijnnaam:
I want to discuss tomorrow during the meeting if the clicky behavior of mac notifications is acceptable or not, and also if the swing notifications are too ugly or not. But, this code is not that complex, so if you want to leave out mac support for now so you can merge this today that's fine with me. (The mac section can just be commented and then this PR should be good to go IMO)

bobismijnnaam:
image

Screenshot for how the TrayIcon notification from Swing looks.

Code:

```
import java.awt.*;
import java.awt.event.*;
import java.net.URL;
import javax.swing.*;

public class Main {


  public static void main(String[] a) throws Exception {
    if (SystemTray.isSupported()) {
      Image image = Toolkit.getDefaultToolkit().getImage("images/tray.gif");
      TrayIcon trayIcon = new TrayIcon(image, "Tester2");

      SystemTray tray = SystemTray.getSystemTray();

      try {
        tray.add(trayIcon);
      } catch (AWTException e) {
        System.err.println("TrayIcon could not be added.");
      }

      trayIcon.displayMessage("Tester!", "Some action performed", TrayIcon.MessageType.INFO);

      Thread.sleep(5000);

      System.out.println("bye");
      System.exit(0);
    }
  }
}
```

niomaster:
Can you quickly try with `UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());`?

niomaster:
I thought I'd just try `SystemTray` myself, but it seems to be unsupported in Ubuntu 20.04 (I guess gnome 3)

niomaster:
Yes, SystemTray support was dropped entirely from gnome 3, but may be enabled again with an addon: https://extensions.gnome.org/extension/1031/topicons/

niomaster:
I recommend against trying this because it adds weird 1 pixel windows and still doesn't look normal with systemlookandfeel

bobismijnnaam:
Too bad, if it is both impractical and ugly we should avoid it. Thanks for looking into it.

bobismijnnaam:
Including the mac look and feel of the above java snippet for completeness:

image

bobismijnnaam:
We discussed this in the meeting and concluded that the current implementation is probably fine. If there is negative feedback about the mac popup's clickability, we'll change the implementation.

hre/src/main/java/hre/util/Notifier.scala
```diff
@@ -0,0 +1,95 @@
...
+    val os = System.getProperty("os.name").toLowerCase()
+    if (os.contains("win")) {
+      notifyWindows10(title, message)
+    } else if (os.contains("mac")) {
+      notifyMacOS(title, message)
+    } else if (os.contains("nix") || os.contains("linux")) {
+      notifyLibnotify(title, message)
+    } else {
+      false
+    }
```
niomaster, line 22: this logic deviates a bit from `Configuration` (but seems slightly better), we should standardize it somewhere.

hre/src/main/java/hre/util/Notifier.scala
```diff
@@ -0,0 +1,95 @@
...
+      |$ToastXml.LoadXml($ToastTemplate.OuterXml)
+      |
+      |$notify = [Windows.UI.Notifications.ToastNotificationManager]::CreateToastNotifier($app)
+      |
+      |$notify.Show($ToastXml)
+      |""".stripMargin
+
+  def notifyWindows10(title: String, message: String): Boolean = {
+    // Script comes from: https://gist.github.com/Windos/9aa6a684ac583e0d38a8fa68196bc2dc
+    val script = powershellNotificationScript.format(title, message)
```
niomaster, line 80: this seems suspect, surely you need some escape here?

niomaster (changes requested):

hre/src/main/java/hre/util/Notifier.scala
bobismijnnaam: I can merge them in a helper method in `Configuration`

hre/src/main/java/hre/util/Notifier.scala
bobismijnnaam: You mean, escaping "title" and "message" for use in powershell, or escaping the powershellNotificationScript variable? Assuming you mean the second case: VerCors doesn't really have a thread model/security concerns afaik, but if we choose to be defensive I'll add an escaping step.

niomaster: I mean the first case, surely if your message is ">" or something it would break.

bobismijnnaam: I guess I'll just whitelist "a-zA-Z!? ", that's easy enough and very effective.

hre/src/main/java/hre/util/Notifier.scala
niomaster: I think I duplicated the logic a few times there, sorry :D

hre/src/main/java/hre/util/Notifier.scala
niomaster: Oh, that is also fine (did not see your reply in time)

niomaster (approved):

Issue #455, opened 2020-06-04
Unresolved name `diz` during type checking


xynxynxyn:
I keep encountering an error in the type checker when using Perm inside loop invariants.
A small example that fails is:
```java
public class Example {
    /*@
       context_everywhere ar != null;
       requires Perm(ar[*], 1);
    */
    public static void example(int[] ar) {
        for (int i = 0; i < ar.length; i++)
        /*@
           requires Perm(ar[i], 1);
        */
        {
            ar[i] = 0;
        }
    }
}
```

which results in the error message

```bash
testing$ vct --silicon Example.java
[warning] (class vct.col.util.FeatureScanner): at file Example.java from line 1 column 7 until line 15 column 0: detected inheritance
[abort] (class vct.col.util.SimpleTypeCheck)At file Example.java from line 7 column 8 until line 13 column 8: unresolved name diz found during type
check at file Example.java from line 7 column 8 until line 13 column 8
The final verdict is Error
```

Operating system: Crux Linux 3.5 x86_64
VerCors version:
```bash
$ vct --version
Vercors 1.3.0
Built by sbt 1.3.8, scala 2.12.10 at 2020-06-04T12:08:52.698Z
On branch dev, commit 024836d,
```

Debug output:
```bash
testing$ vct --silicon Example.java --debug vct.main.Main
[progress] parsing inputs...
[progress] Parsing java file Example.java
[progress] first parsing pass took 200ms
[progress] AST conversion took 70ms
[progress] Flattening variables took 20ms
[progress] Shuffling specifications took 20ms
[progress] post processing took 10ms
[progress] rewriting with/then blocks took 10ms
[progress] interpreting annotations took 10ms
[progress] filtering spec_ignore took 10ms
[progress] Read Example.java succesfully
[progress] Parsed 1 file(s) in: 350ms
[warning] (class vct.col.util.FeatureScanner): at file Example.java from line 1 column 7 until line 16 column 0: detected inheritance
[progress] [01%] java_resolve took 10 ms
[progress] [02%] standardize took 10 ms
[progress] [03%] java-check took 10 ms
[progress] [05%] check took 10 ms
[progress] [06%] infer_adt_types took 10 ms
[progress] [07%] check took 0 ms
[progress] [09%] standardize took 10 ms
[progress] [10%] java-check took 0 ms
[progress] [11%] pointers_to_arrays took 10 ms
[progress] [12%] java-check took 10 ms
[progress] [14%] array_null_values took 10 ms
[...]

PR #552, opened 2020-10-02
Make loop methods static if a loop is in a static method

Drevanoorschot:
This should close issue #455. I'm not too sure whether this is the best solution, particularly because I feel like it only fixes the particular example in the issue.

### DIFF
--- a/examples/basic/StaticLoop.java
+++ b/examples/basic/StaticLoop.java
@@ -1,0 +1,19 @@
+// -*- tab-width:4 ; indent-tabs-mode:nil -*-
+//:: cases StaticLoop
+//:: tools silicon
+//:: verdict Pass
+public class StaticLoop {
+    /*@
+       context_everywhere ar != null;
+       requires Perm(ar[*], 1);
+    */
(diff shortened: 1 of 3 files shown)


bobismijnnaam:
Is your feeling coming from only fixing it for `static`, but there might be other cases/keywords that break iteration contracts? We can consider adding a sort of flag+annotation whitelisting, such that we only allow methods that are either plain or static, and error out on anything else. That way at least we know it works for those two cases, and if someone runs into other cases or we add extra flags/annotations and forget to add it to the whitelist, we'll be notified.

Might be a bit overkill, and your solution looks correct. But it's maybe worth considering if there are enough other cases we're unsure about (@niomaster?).

For completeness, the whitelist would probably be: plain, static, final, ghost, inline (this one I'm unsure about) and the blacklist: pure, extern (as an extern method cannot have a body I guess?). Especially inline methods are maybe a bit weird...

raulmonti:
@Drevanoorschot you were going to keep on working on this one? or did I misunderstood on Tuesday?

Issue #223, opened 2019-06-12
Add option<T> access methods


Sophietje:
Currently an option<T> cannot be unpacked in __PVL__. It would be nice to implement methods to access/unpack values that are wrapped in an option type. In Scala you are able to use `getOrElse`, `forEach` or `match`.

See https://alvinalexander.com/scala/best-practice-option-some-none-pattern-scala-idioms for more information on how options can be unpacked in Scala.

wytseoortwijn:
The axiomatic domain of option types is currently defined in `config/prelude.sil`, as follows:

```
domain VCTOption[T] {
  function VCTNone(): VCTOption[T]
  function VCTSome(t: T): VCTOption[T]
  function getVCTOption(o:VCTOption[T]) : T
    
  axiom not_equal_vct {
    forall x: T :: VCTNone() != VCTSome(x)
  }
  
  axiom get_axiom_vct {
    forall x: T :: getVCTOption(VCTSome(x)) == x
  }
}
```

So there is already a constructor `getVCTOption` for "unpacking" optional values, but I am not sure what the PVL equivalent of this constructor is.

Does PVL currently have a function that translates to `getVCTOption`?

niomaster:
We want to map some operator to `StandardOperator.OptionGet`, from there it is indeed translated to a viper function. Nothing maps to `OptionGet` yet in PVL (or any other language), so we just need to decide on a syntax. Any suggestions? Or do we want to go the functional route, as Sophie suggests?

OmerSakar:
After a discussion, it was decided that we will go for the following syntax: `o1.get()` and `o1.getOrElse(defaultValue)`

OmerSakar:
After some more discussion, it was decided to leave the syntax mentioned in the previous comment in favour of `getOption(o1)` and `getOrElseOption(o1,default)`.

PR #555, opened 2020-10-12
A get function for options 

OmerSakar:
This PR solves issue #223 (partially).

In the issue there were suggestions, but it was never decided which suggestion to implement. To get the ball rolling, I have extended the VCTOption domain to also have a  "getOrElse" function (see the associated axioms below).

Currently, I just extended the syntax (through the PVLSyntax class) to include two functions "optionGet(o1)" and "optionGetOrElse(o1, default)" corresponding to the get operation and getOrElse operation.

```
axiom get_or_else_axiom_1 {
    forall val: T, default: T :: {getVCTOptionOrElse(VCTSome(val), default)} 
        getVCTOptionOrElse(VCTSome(val), default) == val
}

axiom get_or_else_axiom_2 {
    forall default: T :: {getVCTOptionOrElse(VCTNone(), default)} 
        getVCTOptionOrElse(VCTNone(), default) == default
}
```

### DIFF
--- a/col/src/main/java/vct/col/ast/syntax/PVLSyntax.java
+++ b/col/src/main/java/vct/col/ast/syntax/PVLSyntax.java
@@ -50,8 +50,8 @@ public class PVLSyntax {
       syntax.addFunction(Old,"\\old");
 
       syntax.addFunction(OptionSome, "Some");
-      syntax.addFunction(OptionGet, "optionGet");
-      syntax.addFunction(OptionGetOrElse, "optionGetOrElse");
+      syntax.addFunction(OptionGet, "getOption");
+      syntax.addFunction(OptionGetOrElse, "getOrElseOption");
 
       syntax.addFunction(MapBuild, "buildMap");
       syntax.addFunction(MapEquality, "equalsMap");
--- a/col/src/main/java/vct/col/ast/syntax/PVLSyntax.java
+++ b/col/src/main/java/vct/col/ast/syntax/PVLSyntax.java
@@ -50,8 +50,8 @@ public class PVLSyntax {
       syntax.addFunction(Old,"\\old");
 
       syntax.addFunction(OptionSome, "Some");
-      syntax.addFunction(OptionGet, "oget");
-      syntax.addFunction(OptionGetOrElse, "ogetOrElse");
+      syntax.addFunction(OptionGet, "optionGet");
+      syntax.addFunction(OptionGetOrElse, "optionGetOrElse");
 
       syntax.addFunction(MapBuild, "buildMap");
       syntax.addFunction(MapEquality, "equalsMap");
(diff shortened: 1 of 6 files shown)


col/src/main/java/vct/col/ast/syntax/PVLSyntax.java
```diff
@@ -50,22 +50,24 @@ public static Syntax get(){
...
+      syntax.addFunction(MapKeySet, "keysMap");
+      syntax.addFunction(MapCardinality, "cardMap");
+      syntax.addFunction(MapValueSet, "valuesMap");
+      syntax.addFunction(MapGetByKey, "getFromMap");
+      syntax.addFunction(MapRemoveKey, "removeFromMap");
       syntax.addFunction(MapItemSet, "itemsMap");
 
-
       syntax.addFunction(TupleFst, "getFst");
       syntax.addFunction(TupleSnd, "getSnd");
```
raulmonti, line 68: why three new lines here?

src/main/java/vct/col/util/AbstractTypeCheck.java
```diff
@@ -1158,6 +1158,15 @@ public void visit(OperatorExpression e) {
         e.setType((Type) ((PrimitiveType) t).firstarg());
         break;
       }
+      case OptionGetOrElse: {
+        if (!tt[0].isPrimitive(PrimitiveSort.Option)) {
+          Fail("first argument is %s rather then an option ", tt[0]);
+        } else if (!tt[1].comparableWith(source(), (Type) tt[0].firstarg())) {
+          Fail("type of the second argument %s does not match the value type of the option %s", tt[1], tt[0].firstarg());
+        }
+        e.setType(tt[1]);
```
raulmonti, line 1167: why not tt[0].firstarg()? let's say that we actually get a value from the Option parameter, then I would rather keep the original type than the one I use as fallback. Of course these may be the same (I am actually not sure if currently there is another possibility) but in the case there is or that it will be, then I would rather stay with the original type. Or am I misunderstanding what is happening here?

col/src/main/java/vct/col/ast/syntax/PVLSyntax.java
OmerSakar: Does not seem to have a reason, will be fixed in next commit.

src/main/java/vct/col/util/AbstractTypeCheck.java
OmerSakar: Hmm I see your point. For example if we have an option `o1` with a float (currently not supported, but suppose), and I say `getOrElseOption(o1, 2)` where 2 is an integer, the type would become int instead of float. I will change this to `tt[0].firstarg()` (in the next commit).

niomaster: Just my 2ct: the correct solution is to pick the least common supertype of `tt[0].firstarg()` and `tt[1]`, e.g. `getOrElseOption(Some(object), null)` should prefer `tt[0].firstarg()` whereas `getOrElseOption(Some(superclassInstance), subclassInstance)` should prefer `tt[1]`. VerCors can't do that though, so `tt[0].firstarg()` is the better solution I think.

PR #556, opened 2020-10-19
Some generic exceptions fixes

bobismijnnaam:
This fixes some general problems I ran into.

- Enable getMessage and printStackTrace
- Add support for exception-wrapping constructors
- Visit catch blocks while typechecking

### DIFF
--- a/src/main/java/vct/col/rewrite/JavaEncoder.java
+++ b/src/main/java/vct/col/rewrite/JavaEncoder.java
@@ -246,6 +246,11 @@ public class JavaEncoder extends AbstractRewriter {
           ArrayList<DeclarationStatement> parameters=gen_pars(m);
           switch(m.kind){
           case Plain:{
+            if (!(m.getContract() == null || m.getContract().isEmpty())) {
+              m.getOrigin().report("info", "This method cannot be automatically inherited by class %s", cl.name());
+              cl.getOrigin().report("error", "Automatic inheritance of method with non-empty contract is not supported");
+              Fail("Automatic inheritance of method with contract not supported");
+            }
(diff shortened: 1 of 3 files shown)


src/main/java/vct/col/rewrite/JavaEncoder.java
```diff
@@ -255,15 +260,18 @@ public void visit(ASTClass cl){
...
-                create.reserved_name(ASTReserved.Super),
-                null,
-                create_method_name("internal", cl.super_classes[0],m),
-                get_names(parameters)));
-            body.add(create.comment("//TODO: folds of chained predicates in post-condition"));
-            res.add(create.method_kind(m.kind, returns, internal_contract, internal_name, parameters, varArgs, body));
+            // We leave body empty, as the method is guaranteed to have no contract
+            // Therefore no extra proof steps (unfolding predicates, checking pre/post conditions for implications)
+            // are not necessary and we can leave the method abstract
+//            BlockStatement body=create.block();
```
raulmonti, line 266: Should we remove this code instead of commenting it?

raulmonti: There is also a sonar cloud recommendation on this file: "Use already-defined constant 'INTERNAL' instead of duplicating its value here." Could you fix that now that you are modifying the file? :D

raulmonti (approved):

src/main/java/vct/col/rewrite/JavaEncoder.java
bobismijnnaam: Fixed in 345a7eb.

PR #567, opened 2020-11-27
New option for using PVLPrinter

petravandenbos-utwente:
This pull request contains:
- small changes to JavaPrinter
- new class PVLPrinter (I copied a lot from JavaPrinter)
- a new option in Configuration for using PVLPrinter in Main (option usage: --session path-to-file)
- small bug fix in Type

### DIFF
--- a/col/src/main/java/vct/col/ast/print/JavaPrinter.java
+++ b/col/src/main/java/vct/col/ast/print/JavaPrinter.java
@@ -624,7 +624,8 @@ public class JavaPrinter extends AbstractPrinter {
       out.println("");
     }
     out.decrIndent();
-    out.lnprintf("}");    
+    out.lnprintf("}");
+    out.println("");
   }
 
   @Override
@@ -638,7 +639,7 @@ public class JavaPrinter extends AbstractPrinter {
         out.lnprintf("");
       }
       for(ASTNode e:ASTUtils.conjuncts(contract.invariant,StandardOperator.Star)){
-        out.printf("invariant ");
+        out.printf("loop_invariant ");
         nextExpr();
         e.accept(this);
         out.lnprintf(";");
@@ -713,7 +714,7 @@ public class JavaPrinter extends AbstractPrinter {
     s.getType().accept(this);
     out.printf(" %s", s.name());
     if (expr!=null){
-      out.printf("=");
+      out.printf(" = ");
       nextExpr();
       expr.accept(this);
     }
@@ -721,6 +722,11 @@ public class JavaPrinter extends AbstractPrinter {
   }
 
   public void visit(Method m){
+    if (m.kind == Method.Kind.Constructor && m.getBody() instanceof BlockStatement) { //empty constructors don't need to be printed
+      if (((BlockStatement) m.getBody()).isEmpty()) {
+        return;
+      }
+    }
     int N=m.getArity();
     Type result_type=m.getReturnType();
     String name=m.getName();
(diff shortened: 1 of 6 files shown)


col/src/main/java/vct/col/ast/print/JavaPrinter.java
```diff
@@ -713,14 +714,19 @@ public void visit(DeclarationStatement s){
...
     }
     if (!in_expr) out.printf(";");
   }
 
   public void visit(Method m){
+    if (m.kind == Method.Kind.Constructor && m.getBody() instanceof BlockStatement) { //empty constructors don't need to be printed
+      if (((BlockStatement) m.getBody()).isEmpty()) {
+        return;
+      }
+    }
```
niomaster, line 729: See <https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.8.9>, we should also check:
* the number of arguments is 0
* the method throws nothing
* there is no other constructor

col/src/main/java/vct/col/ast/print/PVLPrinter.java
```diff
@@ -0,0 +1,1532 @@
...
+        }
+        out.printf(")");
+        if(i==N-1){
+          out.lnprintf("];");
+        } else {
+          out.lnprintf(",");
+        }
+      }
+      out.lnprintf("  @ * /");
+    } else {*/
```
niomaster, line 825: let's remove this

col/src/main/java/vct/col/ast/print/PVLPrinter.java
```diff
@@ -0,0 +1,1532 @@
...
+                nextExpr();
+                s.getGuard(i).accept(this);
+                out.lnprintf(") ");
+            }
+            s.getStatement(i).accept(this);
+            if (!self_terminating(s.getStatement(i))){
+                out.lnprintf(";");
+            }
+        }
+        //}
```
niomaster, line 852: idem

col/src/main/java/vct/col/ast/type/Type.scala
```diff
@@ -24,7 +24,7 @@ abstract class Type(val args:List[ASTNode]) extends ASTNode {
   def argsJava = args.asJava
   
   /** Tests whether `args` has any arguments. */
-  def hasArguments = args.isEmpty
+  def hasArguments = !args.isEmpty
```
niomaster, line 27: Please use `args.nonEmpty`

src/main/java/vct/main/Main.java
```diff
@@ -329,6 +329,10 @@ public static void main(String[] args) throws Throwable
       } else if (silver.used()||chalice.get()) {
         passes=new LinkedBlockingDeque<String>();
 
+        if(Configuration.session_file != null) {
```
niomaster, line 332: I think this should be `Configuration.session_file.get()`?

niomaster (changes requested):

col/src/main/java/vct/col/ast/print/JavaPrinter.java
petravandenbos-utwente: In method ```public void visit(Method m)``` I do not have information about other constructors. Do you want me to remove this if-statement, and just change back to the situation before this PR?

niomaster: I'm happy to assist in getting this to work better, but you can decide for yourself if it's worth it. With #541 we can probably also ensure that your formatting can happen before the default constructor is added instead.

petravandenbos-utwente: For now, I change it back to what it was before.

niomaster (approved):

PR #571, opened 2021-01-07
Ignore MacOS specific OS file in git

Sophietje:
Added .DS_Store to the gitignore, to ignore these MacOS specific files which contain metadata about display options of folders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ tmp.*
 vct-tool.jar
 *~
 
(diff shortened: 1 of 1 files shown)


niomaster:
Sure! I learned a while back that it's better practice to put local stuff in your personal .gitignore file, see here: <https://stackoverflow.com/a/5724484/871836>, but since we already have a bunch we can add it.

PR #573, opened 2021-01-26
Par encoder refactor

niomaster:
This is an old attempt at cleaning up the parallel block encoder. Perhaps we will continue to work on it some day...


niomaster:
ParallelBlockEncoder is being refactored in new-col now.

PR #574, opened 2021-01-26
Floats

niomaster:
An attempt at increasing support for floats.

### DIFF
--- a/examples/basic/floats.c
+++ b/examples/basic/floats.c
@@ -1,0 +1,10 @@
+void calculations(float a, float b) {
+    float add = a + b;
+    float sub = a - b;
+    float prod = a * b;
+    float div = a / b;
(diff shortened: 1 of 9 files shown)


raulmonti:
Hi @niomaster, I think I recall this axiomatisation to be unsound :( Maybe we would be better of by just compiling our Floats to fracs in Viper, otherwise we should try to fix the axiomatisation but that may take longer. I will see if I have some time to give it a second attempt. Did you change anything in the axiomatisation or is this how it was left?

niomaster:
I don't think we're likely to merge this soon, the recent influx of draft PRs is mostly to document the purpose of floating branches in the repository in a few sentences.

niomaster:
At least superseded by #734

PR #577, opened 2021-01-26
Reinstate carbon

bobismijnnaam:
This PR reinstates the carbon back end. It also splits a test that verifies for carbon but not for silicon into two, namely the KeYAbruptVerificationChallenge example.


bobismijnnaam:
Carbon is mostly functional at this point. It only does not seem to work properly with the test suite yet. This might have to do with running parallel instances of carbon, not sure. It still needs some investigating. But running single examples should work.

bobismijnnaam:
I have not been able to reproduce the parallel tests problem, so it seems that this branch is ready for review and merging.

bobismijnnaam:
When reviewing, be sure to filter out .pdb and .dll using the built-in github filter for readability.

examples/abrupt/KeYAbruptTerminationChallengeSilicon.java
```diff
@@ -0,0 +1,40 @@
+// -*- tab-width:4 ; indent-tabs-mode:nil -*-
+//:: cases KeYAbruptTerminationChallengeSilicon
+//:: tools silicon
+//:: verdict Fail
```
niomaster, line 4: I assume this is expected?

hre/src/main/java/hre/config/Configuration.java
```diff
@@ -297,6 +303,53 @@ public static MessageProcessEnvironment getSilicon() throws IOException {
         return env;
     }
 
+    public static String getMonoVersion() {
+        MessageProcessEnvironment env = new MessageProcessEnvironment("mono");
+        env.addArg("--version");
+        MessageProcess mp = env.startProcess();
+        while (!mp.isFinished()) {
+            Output("Waiting for termination of \"mono --version\"...");
```
niomaster, line 311: uh, what's up with this busy-waiting loop?

hre/src/main/java/hre/config/Configuration.java
```diff
@@ -297,6 +303,53 @@ public static MessageProcessEnvironment getSilicon() throws IOException {
...
+                return matcher.group(1);
+            }
+        }
+
+        return null;
+    }
+
+    public static void checkCarbonRequirements() {
+        String os = System.getProperty("os.name").toLowerCase();
+        if (os.contains("mac") || os.contains("nix") || os.contains("linux")) {
```
niomaster, line 328: Isn't this supposed to work with getOS now?

hre/src/main/java/hre/io/MessageProcessEnvironment.java
```diff
@@ -87,4 +86,15 @@ public MessageProcess startProcess() {
 
     return new MessageProcess(workingDirectory, argv.toArray(new String[0]), env);
   }
+
+  public String getInvocation() {
+    StringBuilder sb = new StringBuilder();
+    sb.append(process);
+    sb.append(" ");
```
niomaster, line 93: this seems suspect, surely strings with spaces should be quoted?

hre/src/main/java/hre/config/Configuration.java
```diff
@@ -297,6 +303,53 @@ public static MessageProcessEnvironment getSilicon() throws IOException {
...
+            // Mono must exist for boogie to run
+            if (!Paths.commandExists("mono")) {
+                Fail("Unable to detect mono. Please ensure the command \"mono\" is available in the PATH.");
```
niomaster, line 331: I think this is a case where "try and ask forgiveness" is better: i.e. just try to obtain the mono version, and output that it's not found if it doesn't work, rather than try to parse `PATH`

niomaster (changes requested):

examples/abrupt/KeYAbruptTerminationChallengeSilicon.java
bobismijnnaam: Yes, it is a shortcoming of the silicon backend. Carbon verifies it just fine

hre/src/main/java/hre/config/Configuration.java
bobismijnnaam: To get the output of "mono --version", we have to wait. Not sure if this can be done asynchronously, as the whole execution of verification and the carbon backend depends on this info. I can insert a sleep(1ms) to make the busy wait less cpu-intensive.

niomaster: I was indeed less worried about the (synchronous) wait-part, and more the busy-part :). You should be able to make a method that uses Process.waitFor in MessageProcess.

hre/src/main/java/hre/config/Configuration.java
bobismijnnaam: Indeed, that needs to be replaced

hre/src/main/java/hre/config/Configuration.java
bobismijnnaam: Good idea

hre/src/main/java/hre/io/MessageProcessEnvironment.java
bobismijnnaam: Since it's only intended for quick debugging of the messageprocessenvironment, and currently unused, I'll delete the method instead.

hre/src/main/java/hre/config/Configuration.java
bobismijnnaam: I refactored it to use raw java processes, using "waitFor" is more efficient and more readable.

PR #578, opened 2021-01-26
Refactor testing to scalatest

bobismijnnaam:
This PR wants to integrate the existing test framework with scalatest. This is desirable because it gives code coverage (JaCoCO) instrumentation for free, as well as being more idiomatic.

However, a downside is that parallelisation does not seem easily possible. It is also unclear if separate runs of examples in the test suite do not influence eachother. Therefore, it is unclear if the scalatest integration can replace the custom test suite, as the custom test suite does have easy parallelisation and an isolation gaurantee.


bobismijnnaam:
After some reflection I don't think this is a good idea, scalatest is good for unit tests but for integration tests the framework we have now is more robust, and not that hard to maintain.

PR #579, opened 2021-01-27
Minimiser

bobismijnnaam:
This PR implements a minimiser routine that minimises the outgoing silver file, based on instructions which methods are of interest. Any methods, functions and fields that are not used are deleted or abstracted iteratively, resulting in a program that contains the minimal proof obligation for the indicated methods.


bobismijnnaam:
I think this PR is almost finished; I was doubting if I should add the feature of being able to specify a line of interest, when then also causes any statements after that line to be deleted, as well as any irrelevant postconditions. But I think that's a bit too much work for marginal gain, so I'll probably merge it as-is (including some final cleaning up).

bobismijnnaam:
Closing this because by now it is too much out of sync with the work done on #629. Maybe in the future we can port over this PR to the new AST.

PR #580, opened 2021-01-27
Upgrade to viper 21.01

bobismijnnaam:
This PR updates:

- Viper to 21.01
- Scala to 2.13.4
- SBT to 1.4.4

To do:

- [x] Test on Linux
- [x] Test on Mac
- [x] Test on Windows
- [x] Change boogie path in carbon script
- [x] Fix the rest of the scripts as well
- [x] Fix benign scala warnings

Testing can be done after compiling using:

```
$ ./bin/vct examples/goto/goto2.pvl --progress --carbon
$ ./bin/vct examples/goto/goto2.pvl --progress --silicon
```

Both runs should pass.


bobismijnnaam:
The code currently doesn't build because of a weird indirect dependency error, probably triggered by some minimal (seemingly irrelevant) code changes on Viper's side and our way of depending on Viper which uses SBT builtins. It can be fixed by a small change on viper's side, see: <https://github.com/viperproject/silver/issues/492>.

build.sbt
```diff
@@ -83,7 +94,6 @@ lazy val vercors = (project in file("."))
     javacOptions in Compile += "-Xlint:deprecation",
     javacOptions in Compile += "-Xlint:unchecked",
     javacOptions in Compile += "-deprecation",
-    javacOptions in doc := Seq(),
```
niomaster, line 86: Does this work? I believe it inherits the options from Compile, hence the empty Seq

src/main/java/vct/logging/VerCorsError.java
```diff
@@ -5,8 +5,8 @@
 
 import hre.ast.Origin;
 import viper.api.ViperError;
-import static vct.logging.VerCorsError.ErrorCode.*;
-import static vct.logging.VerCorsError.SubCode.*;
+//import static vct.logging.VerCorsError.ErrorCode.*;
```
niomaster, line 8: drop these lines?

src/main/java/vct/main/Main.scala
```diff
@@ -176,7 +176,7 @@ class Main {
         silver.set("silicon")
       case "silicon" =>
       case "carbon" =>
-        Configuration.checkCarbonRequirements()
+        // Nothing to check for
```
niomaster, line 179: This suggests that the silicon case falls through to the carbon case, which is not true. Please add a similar comment to silicon or remove it :)

viper/src/main/scala/viper/api/SiliconVerifier.scala
```diff
@@ -7,17 +7,18 @@ import hre.ast.OriginFactory
...
     val silicon = new viper.silicon.Silicon(PluginAwareReporter(HREViperReporter()), Seq("startedBy" -> "example", "fullCmd" -> "dummy"))
 
-    val z3Config = '"' + z3Settings.asScala.map{case (k, v) => s"$k=$v"}.mkString(" ") + '"'
+    val z3KeyValues = z3Settings.asScala.map{case (k, v) => s"$k=$v"}.mkString(" ")
 
     var siliconConfig = Seq(
       "--z3Exe", z3Path.toString,
-      "--z3ConfigArgs", z3Config,
+      // Ensure key-value pairs are wrapped in quotes without leading or trailing spaces
+      "--z3ConfigArgs", s""" "${z3KeyValues}" """.trim,
```
niomaster, line 21: what does this change do? as I see it:

* `s` means we have a format string on our hands
* triple quote means we can put literal quotes in the string
* the string thus is `<space><quote>z3KeyValues<quote><space>`
* we trim it, leaving us with `<quote>z3KeyValues<quote>`

why not just surround z3KeyValues in quotes normally? I think `mkString` accepts a start and end.

niomaster (changes requested):

build.sbt
bobismijnnaam: After removing that line this is what I get in sbt:

```sbt
[IJ]show doc / javacOptions
[info] hre / doc / javacOptions
[info] 	List()
[info] col / doc / javacOptions
[info] 	List()
[info] viper_api / doc / javacOptions
[info] 	List()
[info] parsers / doc / javacOptions
[info] 	List()
[info] doc / javacOptions
[info] 	List()
```

So looks like it was changed in one of the SBT versions. At least the "stage" task and such still work.

viper/src/main/scala/viper/api/SiliconVerifier.scala
bobismijnnaam: Perfect! Although the "\"" still bother me, but I guess that's not avoidable.

PR #582, opened 2021-02-04
Windows support

ArmborstL:
Improve support for VerCors on Windows


hre/src/main/java/hre/config/Configuration.java
```diff
@@ -276,6 +276,7 @@ public static MessageProcessEnvironment getThisVerCors() throws IOException {
         env.addArg("-Xss128M");
         env.addArg("-cp", System.getProperty("java.class.path"));
         env.addArg("vct.main.Main");
+        env.setEnvironmentVar("TEMP", System.getenv("TEMP"));
```
niomaster, line 279: Can you guard this with `if(System.getenv("TEMP") != null)` please? I think the value is not defined in linux, so it will end up passing `TEMP=null` (the string "null").

niomaster (approved):

Issue #341, opened 2019-11-25
Remove options that can be replaced with FeatureScanner


niomaster:
The list of specified/needed options by examples is:

 * --explicit
 * --check-history
 * --check-defined
 * --inline
 * --check-axioms
 * --infer-modifies

We should try to infer those from the AST where possible (though I suspect that's just --check-history)

niomaster:
fixed as of option refactor

Issue #128, opened 2018-09-26
--explicit flag


sjcjoosten:
The verification of:
```
./unix/bin/vct --silicon --explicit ./examples/manual/functions.pvl
```
does not work with the ``explicit`` flag, but the error message is too cryptic

sjcjoosten:
This seems to be an issue with the viper printer. If I write:

```
./unix/bin/vct --silicon ./examples/manual/functions.pvl
```

it simply works, adding the ` --explicit` flag causes a Java exception that should be fixed (probably by generating a unique name somewhere). Here is a minimal example for reproducing this behavior:

```
class counter {
  resource state()=true;
  
  requires state();
  boolean get()=unfolding state() in true;
}
```

niomaster:
The vercors source mentions that explicit exists, because chalice cannot represent predicates with arbitrary arguments. Since we no longer want to support that backend (and silicon can represent predicates with arbitrary arguments), I nominate we remove the functionality altogether.

Issue #79, opened 2018-05-01
Flags cleanup


sjcjoosten:
Vercors has many flags. This issue is there to reduce the number of flags you get when using -h.
Ideally, this would be done by combining similar options into single flags, and cleaning up unnecessary ones.
Finally, we could reduce the number of flags printed when using -h, and have a separate flag for printing all flags.

(Should be assigned to Fauzia)

FauziaE:
The file `histories/history.pvl` mentions that `--check-defined` and `--check-axioms` are disabled due to refactoring. We could perhaps just remove them.

wytseoortwijn:
I would say that the `--silicon` flag should be default. If no back-end is selected when running VerCors, then the default option of using Silicon should be used.

FauziaE:
The following flags seems to be related to kernel examples
---single-group ( used in example  arrays/forward-kernel.pvl)
--disable-auto-barrier
--disable-resource-check
--disable-post-check
--global-with-field |  may also be something related to kernel examples [Verification of Concurrent Systems with VerCors](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.433.3794&rep=rep1&type=pdf)
@sjcjoosten can you please look into these as I think you have worked on kernel examples?

wytseoortwijn:
The `--tool` flag is only used for the testing framework. For example, the command `./unix/bin/vct --tool=silicon --test=examples/arrays` will verify all examples in the `examples/arrays` directory that use Silicon as back-end. We should therefore not remove this flag.

wytseoortwijn:
I think we should attempt to categorise flags, so that `vct --help` shows all flags under a specific category (e.g. testing framework, kernels, futures, back-ends, etc.).

niomaster:
Flags have been re-made in `scopt`.

PR #586, opened 2021-02-08
Update options of VerCors

Sophietje:
- Removes Chalice backend
- Removes Dafny backend
- Removes Silver (silicon_qp) backend
- Removes Boogie backend
- Removes --infer-modifies option
- Removes --explicit option
- Removes --inline option
- Removes --save-intermediate option
- Updates description for --help
- Updates description for --check-defined, --check-axioms and --check-history

Related to isssues
- #428 (because this PR removes unused backends)
- #128 (--explicit option removed as suggested in the last comment by @niomaster)
- #341 (--explicit, --inline, --infer-modifies have been removed since they are no longer used so they don't need to be replaced with FeatureScanner)
- #79


Sophietje:
SonarCloud complains that that I should use constants instead of the strings like "standardize", however this wasn't modified in this PR. So for now, I'll ignore the 7 code smells reported by SonarCloud.

raulmonti:
@niomaster are you solving the merge conflict here?

niomaster (approved):

PR #591, opened 2021-02-24
support given/yields for kernels, support multiple upper bounds in quantified relational expressions

niomaster:


raulmonti:
@Safari1991 are you looking into this PR? you can merge if the solution works for you

Safari1991:
It seems changes in this branch make this example to give an error:

#include <cuda.h>

////////////////////////////////////////////////////////////////////////////////
//Kernel
////////////////////////////////////////////////////////////////////////////////
/*@ 
requires opencl_gcount == 1;
requires \pointer_index(output, \ltid, 1);
@*/
__global__ void CUDA_Kernel_Blelloch(int* output)
{
}
///////////////////////////////////////////////////////////////////////////////

If we change \ltid to \gtid, it passes.

PR #616, opened 2021-03-23
changes to PVL and JavaPrinter and PVLSyntax

petravandenbos-utwente:
This pull-request adds the changes for PVLPrinter.java, JavaPrinter.java and PVLSyntax.java that I made in the session-generation branch


col/src/main/java/vct/col/ast/print/JavaPrinter.java
```diff
@@ -1440,17 +1462,17 @@ public void visit(VariableDeclaration decl){
       sep=",";
       if (dd instanceof DeclarationStatement){
         DeclarationStatement d = (DeclarationStatement)dd;
-        d.getType().accept(this);
+        out.print(d.name());
```
bobismijnnaam, line 1463: The type of the variable declaration is now no longer printed, are you sure that is correct? Or is the "decl.basetype.accept(this)" line above now responsible for printing the type?

col/src/main/java/vct/col/ast/print/PVLPrinter.java
```diff
@@ -598,23 +597,32 @@ public void visit(Contract contract) {
                 e.accept(this);
                 out.lnprintf(";");
             }
-            for(ASTNode e:ASTUtils.conjuncts(contract.pre_condition,StandardOperator.Star)){
-                out.printf("requires ");
-                nextExpr();
-                e.accept(this);
-                out.lnprintf(";");
+            List<ASTNode> contextElems = new ArrayList<>();
```
bobismijnnaam, line 598: Since this implementation is the same as in JavaPrinter, maybe put this in abstractprinter, the superclass? Then the JavaPrinter only has to put the /*@ @*/ around I think.

col/src/main/java/vct/col/ast/print/PVLPrinter.java
```diff
@@ -652,6 +660,17 @@ public void visit(Contract contract) {
         }
     }
 
+    private void printContractElement(ASTNode expr, String contractHead) {
```
bobismijnnaam, line 661: This one is also basic and generic, canyou put this in AbstractPrinter?

bobismijnnaam (changes requested):
Some minor suggestions, please have a look if you think they are possible. Besides that it looks good for merging!

col/src/main/java/vct/col/ast/print/JavaPrinter.java
petravandenbos-utwente: > The type of the variable declaration is now no longer printed, are you sure that is correct? Or is the "decl.basetype.accept(this)" line above now responsible for printing the type?

Yes "decl.basetype.accept(this)" is now responsible for printing the type.

col/src/main/java/vct/col/ast/print/PVLPrinter.java
petravandenbos-utwente: Done.

OmerSakar:
One more change that could be part of this PR. 

Currently, the JavaSyntax is taken as the "diagnostic" language syntax. Given that COL and PVL are more closely related, it would be nice to change it to PVLSyntax. This will then print the COL AST using the PVLSyntax (e.g. when we use the --show-after or --show-before flags).

The diagnostic language syntax is set in the file `vct/col/ast/util/Configuration.scala` (see the line below).

```
def getDiagSyntax: Syntax = JavaSyntax.getJava(JavaDialect.JavaVerCors)
```

col/src/main/java/vct/col/ast/print/PVLPrinter.java
petravandenbos-utwente: visit(Contract contract) is now a method in AbstractPrinter, and JavaPrinter adds the @'s and indents to it.

petravandenbos-utwente:
> One more change that could be part of this PR.
> 
> Currently, the JavaSyntax is taken as the "diagnostic" language syntax. Given that COL and PVL are more closely related, it would be nice to change it to PVLSyntax. This will then print the COL AST using the PVLSyntax (e.g. when we use the --show-after or --show-before flags).
> 
> The diagnostic language syntax is set in the file `vct/col/ast/util/Configuration.scala` (see the line below).
> 
> ```
> def getDiagSyntax: Syntax = JavaSyntax.getJava(JavaDialect.JavaVerCors)
> ```

By changing it to PVL, I get an error for examples/parallel/array_par.pvl, so some things are missing...
[debug] java.lang.Error: no syntax defined for Subscript operation

Sophietje:
It'd be great if the SonarCloud code smells can be fixed before merging

Issue #631, opened 2021-05-03
Better management of the .classpath file


bobismijnnaam:
The `bin/.classpath` file can currently get out of sync with the project.

To make the interface more intuitive, `bin/.classpath` should be deleted when the sbt `clean` task is executed. This is already in `build.sbt`, but because of a bug sbt doesn't actually delete the file on clean. This should be fixed in sbt 1.5, so upgrading sbt will improve the situation already.

To completely fix it, I propose that the `compile` action of sbt also writes to `bin/.classpath`. I think this can be achieved by overriding the `compile` action in `build.sbt` to also write to the `bin/.classpath` file.

niomaster:
The classpath is computed from the build files, and sbt only updates the build files on reload. It would make more sense to hook the .classpath update to reload and/or startup then.

bobismijnnaam:
Yes, that's also okay, as long as compilation also updates it (e.g. when adding classes or subprojects the .classpath might need to be updated?). Specifically, the ideal behaviour I would like is: if you've compiled vercors since you've made changes or added any of the source files, the `bin/.classpath` file is up to date.

niomaster:
The types of changes you list sketch exactly the distinction between compilation an reloading:

- "adding classes ~or subprojects~": if sbt compiles them, they are guaranteed to be under a classpath that sbt already knows about, and hence `.classath` already knows about. Compilation is guaranteed not to change the classpath.
- "~adding classes~ or subprojects": sbt cannot load or change out new (sub)projects while running, and will not compile any new code in a new project, or drop projects that are moved/deleted. Only when you reload the project, or restart sbt entirely, sbt knows about the new project. Correspondingly, the class path is not updated if you change the project structure while sbt is running, so it will not help to update `.classpath` after a compile in that case.

bobismijnnaam:
Interesting, then start/reload indeed seems to be all we need.

Issue #603, opened 2021-03-12
Script to remove all target-stuff


Sophietje:
It seems to happen quite often, when someone pulls vercors for updates, that (s)he needs to remove all target directories to get a correct compilation. Would it be possible to either make a script (or include in `sbt clean`?) so that we don't need to manually remove these files?

Sophietje:
This should perhaps also include bin/.classpath

bobismijnnaam:
From <https://www.scala-sbt.org/1.x/docs/Running.html#Common+commands>: 

> | Command | Description |
> |-- | --|
> |clean | Deletes all generated files (in the target directory). |

I'd say that clean should fix our problem but apparently not...?

bobismijnnaam:
One clue: I noticed today that running clean only triggers recompilation of the main project ("vercors"), and not recompilation of the sub projects, nor the compiled class files of `build.sbt`. Might be relevant.

Or we can just roll our own clean target that deletes every "target" folder in the project. But we need to be sure that no-one ever creates a file called "target" if we do that...

niomaster:
It's because our dependencies are not `aggregate`'ed

bobismijnnaam:
So we have to roll our own because we fall outside the "standard" sbt mold?

niomaster:
We can probably aggregate them instead, though I vaguely recall a reason why we do not do it. We should investigate it again.

bobismijnnaam:
I applied the fix on PR https://github.com/utwente-fmt/vercors/pull/580, locally the building seems to work fine. From [the docs](https://www.scala-sbt.org/1.x/docs/Multi-Project.html#Aggregation) it seems to do exactly what we want, and it also actually deletes code.

The only line that bothers me is the following:

> Note: aggregation will run the aggregated tasks in parallel and with no defined ordering between them. 

I guess this might be problematic since we generate the parsers, so if that happens late the build will fail. Unfortunately I have a hard time finding more documentation about aggregation (I'm also surprised why `dependsOn` and `aggregate` are separate...?)

I'll do a few more test runs with different viper commits as dependencies, and the CI will have a stab at it as well.

bobismijnnaam:
The easiest path forward is by adding the `.classpath` file to the `clean` task. This is straightforward, except for a tiny bug in sbt (sbt/sbt/issues/6395) (I think), so pending that we can add it if everyone else agrees it's a satisfactory workaround.

niomaster:
[...]

PR #632, opened 2021-05-04
Upgrade SBT to 1.5.1

bobismijnnaam:
Upgrade sbt to 1.5.1, specifically to include a fix in SBT needed for ~#603~ #631.

Before merging:
- [x] Fix the warnings introduced by deprecated syntax in 1.5


Vescatur (approved):

Vescatur:
Does this also solve issue #631? In case yes, could you close the issue?

bobismijnnaam:
It does help with better management of `.classpath`, as the new sbt will actually delete it when you run `sbt clean`. But besides that it doesn't improve the situation I think.

PR #642, opened 2021-06-03
COL Redesign: Issue 637: unittest FilterSpecIgnore in the new AST

### LINKED ISSUE #637: Create a unit test for FilterSpecIgnore Pass

Vescatur:
This merge adds helper classes for writing pass tests.
Two tests have been added for the pass FilterSpecIgnore.
Two test which expect an error have been added but are not yet finished. The expected error of FilterSpecIgnore has not yet been implemented. For now the test throws an NotImplementedError. #closes #637


Vescatur:
This is a branch to merge issue-637 to ast. The test do not yet work in the ast therefor the test are failing. The code smells is comparing issue-637 to dev which is why there are so many. This pull request only adds one which is a TODO.

col/build.sbt
```diff
@@ -19,21 +19,17 @@ generateHelpersTask := {
     src / "C.scala",
   )
 
-  /*
-  val compile = FileFunction.cached(streams.value.cacheDirectory / "antlr4", FilesInfo.hash, FilesInfo.hash)(changedSet => {
+
+  val compile = FileFunction.cached(streams.value.cacheDirectory / "removeThisToGenerate-src_managed", FilesInfo.hash)(changedSet => {
```
Vescatur, line 23: This code will check if the folder "removeThisToGenerate-src_managed" is changed. If that is the case it will regenerate the Scala classes.

col/src/main/java/vct/col/ast/Rewriter.scala
```diff
@@ -1,8 +1,17 @@
...
 
+import vct.col.ast.RewriteHelpers._
+
 class Rewriter extends AbstractRewriter {
   override def dispatch(stat: Statement): Statement = rewriteDefault(stat)
-  override def dispatch(e: Expr): Expr = rewriteDefault(e)
+  override def dispatch(e: Expr): Expr = {
+    e match {
+      case node: Constant.BooleanValue =>
+        new Constant.BooleanValue(node.value)(node.o)
```
Vescatur, line 10: This was added because the automatically generated class does not generate methods for BooleanValue.

bobismijnnaam: That sounds like a bug in the generation code. Can ensure that the methods are generated, or is there something preventing that?

niomaster: it's definitely a bug, I intend to have every node auto-generated, just haven't looked at it yet.

col/build.sbt
bobismijnnaam: Ah, right, I see why it was changed from antlr now. Is the approach now automatic or do you have to manually remove the folder to trigger regeneration of the class files?

Vescatur: To be honest. I am not that familiar with build.sbt files. So I think you have to manually remove it. I am not sure what the problem was with an automatic approach.

col/src/main/java/vct/col/ast/Rewriter.scala
bobismijnnaam: Ah, is this because IntegerValue and BooleanValue are nested with the Constant object, instead of top-level case classes?

col/build.sbt
niomaster: This is in fact automatic, it is just incidentally also cached. If for whatever reason the caching mechanism fails, you can indeed *also* remove this directory to trigger re-generation.

col/src/main/java/vct/col/ast/Rewriter.scala
niomaster: Good guess! I added support for arbitrary nesting of `object`'s though, so potentially not, but it's probably related (as it seems to be the only one not generated)

bobismijnnaam: I think this piece of code needs to be added to collectBases:
```
    case Defn.Object(_, _, Template(_, _, _, stats)) =>
      stats.foreach(collectBases(_))
```
@niomaster Should I create a PR for this? (Just making sure I'm not duplicating work)

niomaster: ah, that seems right. Perhaps also `collectFamily`?

bobismijnnaam: Sure, I'll create a PR.

col/build.sbt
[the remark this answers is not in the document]
Vescatur: No, the cached directory is src_managed. Which should automatically regenerate if the code changes. However if the automatic system fails, you can remove the directory "removeThisToGenerate-src_managed".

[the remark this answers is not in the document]
bobismijnnaam: We could also call it "removeThisToGenerate-COLHelpers"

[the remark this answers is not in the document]
bobismijnnaam: Ah wait, the "removeThisToGenerate" folder is only an indicator folder, dwz it doesn't contain anything useful but is only there to allow you to trigger regeneration?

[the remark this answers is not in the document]
Vescatur: Yes, indeed

src/test/scala/vct/helper/SimpleProgramGenerator.scala
```diff
@@ -0,0 +1,26 @@
+package vct.helper
+
+import vct.col.ast.Constant._
+import vct.col.ast.{ApplicableContract, Class, FileOrigin, InputOrigin, InstanceMethod, Origin, Program, Statement, TVoid}
```
bobismijnnaam, line 4: Utility code like this is fine, but it seems pretty specific for each unit test. So maybe this can instead be included in the unit tests class, or in the same file but as a separate object? Then you can also pick better names, for example, instead of generateSimpleInputOrigin, you can just use generateOrigin, since it's usage locally in the unit test class is clear. ("Simple" is a very bad name anyway, imo :) )

src/test/scala/vct/newrewrite/FilterSpecIgnoreSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import vct.col.ast._
+import vct.col.newrewrite.FilterSpecIgnore
+import vct.helper.{AstComparer, RewriteTestHelper, SimpleProgramGenerator}
+
+
+class FilterSpecIgnoreSpec extends AnyFlatSpec with Matchers {
+
+  //nothing should change
```
bobismijnnaam, line 12: If these comments are implemented, they can be removed right? Or do they still need to be done? IN that case they can be made TODO or can be converted into github issues?

src/test/scala/vct/newrewrite/FilterSpecIgnoreSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+      val body = Block(Seq(
+        LocalDecl(variable),
+        Return(Local(new LazyRef[Variable](variable)))
+      ))
+      programExpectedOutput = SimpleProgramGenerator.generateProgramWithSingleClassAndSingleMethod(body)
+    }
+    RewriteTestHelper.test(rewriter, programInput, programExpectedOutput)
+  }
+
+  it should "throw error with two many SpecIgnoreStart" in {
```
bobismijnnaam, line 75: It should be "too many", also a few times below.

src/test/scala/vct/helper/SimpleProgramGenerator.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  def generateSimpleInputOrigin(): InputOrigin ={
+    FileOrigin(Paths.get(""), 1, 1, 1, 1)
+  }
+
+  def generateSimpleApplicableContract()(implicit origin: Origin = generateSimpleInputOrigin()): ApplicableContract ={
+    ApplicableContract(BooleanValue(value = true),BooleanValue(value = true),BooleanValue(value = true),Seq(),Seq(),Seq())
+  }
+
+  def generateProgramWithSingleClassAndSingleMethod(body: Statement)(implicit origin: InputOrigin = generateSimpleInputOrigin()): Program ={
```
bobismijnnaam, line 18: If you make this method local, you can call it just "procedure" or "method", like Pieter did in the Viper unit tests as well

bobismijnnaam (commented):
General code is good but there are a few finishing touches I'd still like to apply. Since this PR is already merged, maybe you can apply the ones you agree with in a new PR?

src/test/scala/vct/helper/AstComparer.scala
```diff
@@ -0,0 +1,119 @@
+package vct.helper
+
+import vct.col.ast.{Constant, Declaration, Node, Program, Ref}
+
+import scala.collection.mutable
+
+
+/**
+ * This object is to compare two programs of the ast.
+ */
```
bobismijnnaam, line 10: Should we add the assumption here about how the comparer works? (Meaning, the first pair of declarations that is encountered is assumed to be equal, or something?)

PR #645, opened 2021-06-14
reduce veymont code smells

petravandenbos-utwente:


src/main/java/vct/col/veymont/Decompose.scala
```diff
@@ -140,27 +136,11 @@ class Decompose(override val source: ProgramUnit) extends AbstractRewriter(null,
...
-              case Some(c) => {
-                if(!c.fields().asScala.forall(_.`type` match{ case p : PrimitiveType => isAllowedPrimitive(p); case _ => false}))
-                  Fail("VeyMont Fail: channel of type %s not supported, because fields are not primitive")
-                cloneClasses = cloneClasses + c
-                result = create.invokation(create.field_name(chanName), null, chanWriteMethodName, create.invokation(sendExpression, null, "clone"))
-              }
-              case None =>
-                Fail("VeyMont Fail: channel of type %s not supported", chanType)
-            }
+            case p : PrimitiveType => checkChanPrimitiveType(p,writeChanName,sendExpression,a)
```
Vescatur, line 142: I think this is really nice. For me it is easier to read.

src/main/java/vct/col/veymont/GenerateLTS.scala
```diff
@@ -196,44 +148,43 @@ class GenerateLTS(override val source : ProgramUnit, isGlobal : Boolean) extends
...
-        case _ => Set.empty
-      }
+      case a : AssignmentStatement => getSubjectAssignment(a)
       case _ : IfStatement => roleNames.toSet
       case _ : LoopStatement => roleNames.toSet
       case p : ParallelRegion => p.blocks.map(_.block).toSet.flatMap(getSubjects(seen,_))
       case m : MethodInvokation => {
         if(m.`object` == null) { //it is a main method
-          Fail("VeyMont Fail: encountered method call %s in LTS generation",m.method)
-          Set.empty
```
Vescatur, line 230: I see we use both Fail and throw Failure. What is our policy for throwing errors. When do we use which one?

Vescatur: Is there a reason you changed it?

Vescatur (changes requested):

src/main/java/vct/col/veymont/GenerateLTS.scala
niomaster: `throw Failure` indicates to scala that the code does not terminate normally (i.e. always results in an exception), so you can use it in place of any value (note that this method returns something). Ideally we want the logging stuff to be written in scala, so we can indicate that `Fail` (the equivalent method) behaves in this way. That's not currently possible because it's written in Java. imo we want to use `Fail` whenever possible, but this usage is better for now in places where a value is expected, and hopefully we can soon use `Fail` in those places also.

Vescatur (commented):
Looks good

Issue #449, opened 2020-06-02
Sonarcloud


niomaster:
In my opinion, Sonarcloud is currently failing in its role as a useful step in our continuous integration, and we need it to change drastically. Please note that I'm not against linting in general, but these issues have become too annoying to leave them be. Some issues from the top of my head:

* Coverage is measured via the sbt test suite, instead of our normal examples test suite. All PRs fail the sonarcloud check because of this
* We haven't encoded local naming conventions properly. For example, StandardOperator uses a different naming convention to regular enum's, causing sonarcloud to fail any PR with a new operator, new ast factory method, etc.
* Linting is too strict when changing a file someone else wrote. "Fixing stuff locally" is fine, but I think we should be careful reformatting large files multiple people may be working on, leading to large merge conflicts (think AbstractRewriter, Main, that sort of thing)
* Sonarcloud does not work in forks, so it fails any PRs from forks. Those typically are merged without checking within the vercors project.

This leads to a situation where in almost any non-trivial pull request Sonarcloud fails the pull request. When deciding whether to merge the conversation usually is "oh it's just sonarcloud complaining? let's just merge it," so we merge pull requests even though the CI fails it. I think this, at a bare minimum, should not be the case.

Here are a few solutions:

* We fix all the sonarcloud issues, leaving it as a requirement for PRs. I think this solution is not viable, as these issues have been long-standing and somewhat "boring," and the cost-benefit ratio is not really there.
* We investigate whether we can make sonarcloud optional, so it shows up as a CI check, but it acts as a warning in that errors do not fail the PR.
* We disable sonarcloud altogether.

I'm curious to hear your opinions!

bobismijnnaam:
Maybe one more solution that can be considered is further tweaking sonarcloud. This would mean properly hooking up our tests to sonarcloud, telling sonarcloud about our naming conventions, and white/blacklisting files/sections of files that we do not want to be linted (took me a while to find, but sonarcloud supports it: https://sonarcloud.io/documentation/project-administration/narrowing-the-focus/). The only real blocker that you mentioned, in my opinion, is that sonarcloud does not work in forks. This could be worked around by only considering pull requests from local branches (something I as a fork user wouldn't mind), but this would still rule out checking forks with sonarcloud.

We could also look at lighter alternatives that are easier to work with or can be run locally. But sonarcloud seems to be more or less the standard at the moment.

bobismijnnaam:
[...]

PR #648, opened 2021-07-07
Issue 449 remove commented code

Vescatur:
This is related to #449 but does not close it.
This branch removes all the commented code. (The branch name is a bit misleading.)


bobismijnnaam:
please keep the comments in the following files: checkprocessalgebra, cslencoder, sequenceutils, and silverbackend. The first three have some significance w.r.t. soundness/features we used to support. SilverBackend contains  some options for viper that are nice to keep around, but if you really want to get rid of the comments maybe you can turn this last one into an actual command line option?

bobismijnnaam:
Thanks 👍

Issue #509, opened 2020-08-12
Run slow suite at regular interval


niomaster:
We should use the travis "cron" feature to regularly build the normally excluded suites of examples, so they don't break (I have no idea if they are currently broken, for example). The suites are: slow,medium,problem-fail,skip-travis

I propose we build them monthly. We should only run the examples in the above suites (not the full test suite), which we can distinguish with `TRAVIS_EVENT_TYPE=cron`

bobismijnnaam:
Another idea: maybe in the regular test suite runs, we can include the excluded suites at least for all pre-viper phases. I.e. forall cases in excluded suites, add the flag ```--stop-before silver``` (or maybe ```--no-check```?). That way typing and other properties of the excluded tests are still checked, but we don't have to wait for the viper phase to complete.

bobismijnnaam:
It has been suggested during the meeting that we can also only run the slow tests on the "dev" branch, since commits there are relatively infrequent anyway.

bobismijnnaam:
Related: #487. But, it's probably better to move the slow/medium suite to a new PR, and (finally) close the coverage PR.

bobismijnnaam:
I removed the weekly/slow suite updated code from #487. If needed the changes that I removed can be found at: 776ade43dfa4d08db548899453028575e0753ef0

PR #649, opened 2021-07-08
Issue 509 slow tests

Vescatur:
This pull request ensures that the slow tests are run each time a pull request is accepted.
In this commit the slow tests are run. https://github.com/utwente-fmt/vercors/runs/2840658941 Note that problem fail test also ran but that part has been removed from the ci.

One of the tests can be fixed by code from Bob. https://github.com/utwente-fmt/vercors/issues/509#issuecomment-868325583 However I did not include this code because the scope of the pull request is to run the slow tests. It is not the goal to fix all the tests. It would also cause me to be the author of code which Bob has wrote. I would suggest that Bob makes a pull request to fix this example.


examples/fixed-known-problems/rewriterIssue4.pvl
```diff
@@ -1,7 +1,7 @@
 // -*- tab-width:2 ; indent-tabs-mode:nil -*-
 
 //:: cases RW4
-//:: suite skip-travis
+//:: suite problem-fixed
```
bobismijnnaam, line 4: Is the problem-fixed suite run on the CI?

bobismijnnaam: Looks like it since only suites are excluded, but just want to doublecheck if you agree

examples/known-problems/verifythis/2019/challenge2b.pvl
```diff
@@ -1,120 +1,120 @@
...
-        }
-      }
-      i++;
-    }
-  }
-}
+// -*- tab-width:2 ; indent-tabs-mode:nil -*-
+//:: cases VerifyThis19TwoB
+//:: tools silicon
+//:: suite medium problem-fail
```
bobismijnnaam, line 4: Why is this entire file changed, even though you seem to only change the suite? It might require some git hacking to drop the unnecessary changes... Maybe the newlines were changed accidentally?

bobismijnnaam (changes requested):
monotonicbool.pvl should work by now, could you double check if does, and if so remove it from the problem-fail suite? The rest looks great, nice work.

examples/known-problems/verifythis/2019/challenge2b.pvl
Vescatur: I am not sure why it is showing it in this way. However git blame in the IDE is working correctly.

examples/fixed-known-problems/rewriterIssue4.pvl
Vescatur: Well actually only RW 1,2,3, and 7 are run. 4, 5 and 6 are missing. I am not sure why.

Vescatur: I found them and they are running in the CI.

Vescatur:
> monotonicbool.pvl

unfortunatly not. This is the output.

[progress] [84%] generateQuantifierTriggers took 31 ms
Pass generateQuantifierTriggers did not remove QuantifierWithoutTriggers:
== Example 1: BindingExpression ==
(\exists int i;0 <= i && i < | xs |;xs [ i ])
=== examples\known-problems\parallel\monotonicBool.pvl ===
  42   seq<boolean> toSeq(boolean[] xs) = toSeqImpl(xs, 0);
  43 
                         [--------------------------------
  44   ensures \result == (\exists int i = 0..|xs|; xs[i]);
                          --------------------------------]
  45   boolean foldOr(seq<boolean> xs) =
  46     |xs| > 0 
  Originating here ^


== Example 2: BindingExpression ==
(\exists int i;0 <= i && i < N;VCTArray<CT:Ref>.loc(VCTOption<T:VCTArray<CT:Ref>>.getVCTOption1(diz.C_contrib),i).Boolean__item)
=== examples\known-problems\parallel\monotonicBool.pvl ===
  83     }
  84 
                    [----------------------------------
  85     assert b == (\exists int i = 0..N; contrib[i]);
                     ----------------------------------]
  86   }
  87 }
  Originating here ^


[progress] [88%] scaleAllPredicateApplications took 15 ms
[progress] [92%] collectInnerDeclarations took 0 ms
[progress] [96%] collectDeclarations took 0 ms
[progress] verifying with builtin silicon backend
[progress] conversion took 78ms
Errors! (1)
=== examples\known-problems\parallel\monotonicBool.pvl ===
  65         ** b == foldOr(toSeq(contrib))
  66         ) {
          [-------------------
  67       par(int tid = 0..N)
  68         context N > 0;
  69         context Perm(contrib, read) ** contrib != null;
  70         context Perm(contrib[tid], 1\2);
  71       {
  72         if (p(tid)) {
  73           // Models: atomic_set(b);
  74           atomic(inv) {
  75             b = true;
  76             contrib[tid] = true;
  77             // Fails:
  78             // b = false;
  79             // contrib[tid] = false;
  80           }
  81         }
  82       }
     -------]
  83     }
  84 
  CallPreCondition:InsufficientPermission
=== examples\known-problems\parallel\monotonicBool.pvl ===
  67       par(int tid = 0..N)
  68         context N > 0;
                         [-------
  69         context Perm(contrib, read) ** contrib != null;
                          -------]
  70         context Perm(contrib[tid], 1\2);
  71       {
  caused by
[progress] task Viper verification took 5608 ms
The final verdict is Fail
[progress] entire run took 16923 ms

Process finished with exit code 0

bobismijnnaam:
Okay, I've handled monotonicBool in the problem-fail PR. I'll also review this PR.

bobismijnnaam (approved):
Looks good. Quite a few new problem-fails we'll have to look at why they are broken...

PR #652, opened 2021-07-08
Fork examples

petravandenbos-utwente:
These examples help to understand how programs with fork and join can be verified.


bobismijnnaam (changes requested):
Looks good, except the indentation in forkjoinforloop.pvl is a bit off. Would you mind changing that?

bobismijnnaam (approved):

PR #654, opened 2021-07-10
COL redesign: Implement drafts of ContinueToBreak and SpecifyImplicitLabels

bobismijnnaam:
- [x] Process Pieter's comments
  - [x] Use scopedstack for labels
  - [x] Use layered origin
- [x] Write tests
  - [x] SpecifyImplicitLabels
  - [x] ContinueToBreak


Vescatur (changes requested):
Could you take a look at why the junit test are failing. I think all of them should pass on the ast branch. 

I noticed there were no unit test written for this new rewrite. I think we should at least have a couple of them per rewrite however that is my personal opinion. I would like to discuss it and make a decision as a team.
This pull request shows how we could make unittests. https://github.com/utwente-fmt/vercors/pull/642

I also want to note that the pull request has 349 code smells. I however think these are mostly falls positives because it is compare the branch ast-continue2break-specifycimplicitlabels with dev. It should compare it with ast. Maybe when sonar has access to the git history it will be more reliable.

Vescatur:
> Could you take a look at why the junit test are failing. I think all of them should pass on the ast branch.

The junit test started failing when I merged dev into ast so it is not because of this pull request. It is still something to look into.

bobismijnnaam:
> we should at least have a couple of them per rewrite

Totally agree, especially since these two passes are very suitable for testing. (For very complicated passes I am not so sure, but maybe we can still test simple properties even then.) I don't think there are any other unit tests in the AST branch for new rewrites yet, but with the new AST having structural equality, this shouldn't be to hard. I'll look into writing tests for the passes in this branch.

bobismijnnaam:
> The junit test started failing when I merged dev into ast so it is not because of this pull request. It is still something to look into.

One failure is because a z3 path is hardcoded, which is probably fixable. The other errors are because of some problem with the new ast that appeared after Pieter made some changes to DirectRef, I think. I have a fix in mind but probably Pieter should look at it first when he is back.

Vescatur:
> > we should at least have a couple of them per rewrite
> 
> Totally agree, especially since these two passes are very suitable for testing. (For very complicated passes I am not so sure, but maybe we can still test simple properties even then.) I don't think there are any other unit tests in the AST branch for new rewrites yet, but with the new AST having structural equality, this shouldn't be to hard. I'll look into writing tests for the passes in this branch.

In the previous mentioned pull request there are some helper functions. I would recommend having a look if those are useful for your case. In that branch there are unit test for the new rewrite "FilterSpecIgnore.scala".

src/main/java/vct/col/newrewrite/SpecifyImplicitLabels.scala
```diff
@@ -0,0 +1,45 @@
...
+    case block@Block(Seq(Label(labelDecl), _: Loop)) =>
+      labelStack.push(labelDecl)
+      val res = rewriteDefault(block)
+      labelStack.pop()
+      res
+    case block@Block(Seq(Label(labelDecl), _: Switch)) =>
+      labelStack.push(labelDecl)
+      val res = rewriteDefault(block)
+      labelStack.pop()
+      res
```
niomaster, line 22: This pattern occurs so much that I usually write a helper for it, see e.g. ScopeContext.collectInScope. Perhaps we should make a generic util version of this somewhere.

src/main/java/vct/col/newrewrite/SpecifyImplicitLabels.scala
```diff
@@ -0,0 +1,45 @@
...
+    case loop: Loop =>
+      implicit val o = loop.o
+      val labelDecl = new LabelDecl()(SourceNameOrigin("loop", o))
```
niomaster, line 25: I intended to use SourceNameOrigin strictly for names that come from source code. I think it's a good idea to implement a custom origin for each pass and/or generation concept, so we can track derivations precisely (and preferably layered, so it describes itself as something like "the label for the loop at $x" or so).

src/main/java/vct/col/newrewrite/SpecifyImplicitLabels.scala
niomaster: It is now `hre.util.ScopedStack`, which you can use like:

```scala
val labelStack = ScopedStack[LabelDecl]()

labelStack.having(labelDecl) {
  rewriteDefault(block)
}
```

src/main/java/vct/col/newrewrite/ContinueToBreak.scala
```diff
@@ -0,0 +1,46 @@
...
+  override def preferredName: String = "continue" + loopLabel.capitalize
+  override def messageInContext(message: String): String = inner.messageInContext(message)
+}
+
+case class ContinueToBreak() extends Rewriter {
+  val loopLabelToInnerLabel = new mutable.HashMap[LabelDecl, LabelDecl]()
+
+  override def dispatch(stat: Statement): Statement = stat match {
+    case block@Block(Seq(l@Label(labelDecl), loop: Loop)) =>
+      val rewrittenBody = rewriteDefault(loop.body)
```
niomaster, line 19: Pretty sure you mean dispatch(loop.body) here

src/main/java/vct/col/newrewrite/ContinueToBreak.scala
```diff
@@ -0,0 +1,46 @@
...
+
+    case c@Continue(Some(Ref(labelDecl: LabelDecl))) =>
+      // Reuse an already created inner label or create one
+      val innerLabelDecl = loopLabelToInnerLabel.get(labelDecl) match {
+        case Some(l) => l
+        case None =>
+          val innerLabelDecl = new LabelDecl()(ContinueToBreakOrigin(labelDecl.o.preferredName, labelDecl.o))
+          loopLabelToInnerLabel.put(labelDecl, innerLabelDecl)
+          innerLabelDecl
+      }
```
niomaster, line 40: I think this implements `Map.getOrElseUpdate`

niomaster: Some food for thought: The name in the silver backend for labels is indeed derived from `LabelDecl` (so this is entirely correct), but this is kind of arbitrary and hard to remember. But maybe this is only confusing for Labels?

src/main/java/vct/col/newrewrite/SpecifyImplicitLabels.scala
```diff
@@ -0,0 +1,41 @@
...
+    case _: Loop => true
+    case _: Switch => true
+    case _ => false
+  }
+
+  override def dispatch(stat: Statement): Statement = stat match {
+    case block@Block(Seq(oldLabel@Label(_), s: Statement)) if isBreakable(s) =>
+      val newLabel = oldLabel.rewrite()
+      val newS = labelStack.having(newLabel.decl) {
+        rewriteDefault(s)
```
niomaster, line 25: dispatch

niomaster (changes requested):

src/main/java/vct/col/newrewrite/SpecifyImplicitLabels.scala
bobismijnnaam: In this case I want the case on line 28 not to trigger if the case on line 22 is triggered, so here I think it _should_ be `rewriteDefault`...?

src/main/java/vct/col/newrewrite/ContinueToBreak.scala
bobismijnnaam: I guess the question here is: can there be two instances of `Label` that share a `LabelDecl`? From a parsing perspective I'd say no...?

PR #656, opened 2021-07-12
Final fix for coverage

bobismijnnaam:
There was a bug that caused the coverage reports not to be merged correctly in the action, causing sonarcloud to display 0% coverage.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -210,6 +210,8 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: vercors-jacoco-xml
+      - name: Debug ls
+        run: ls -alh jacoco*
       - name: SonarCloud Scan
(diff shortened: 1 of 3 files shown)


bobismijnnaam:
I think the coverage code smell can be ignored, as it concerns coverage in the test suite code, and it requires changing the CI to get the coverage there, so I don't think it's worth the effort.

Vescatur:
> I think the coverage code smell can be ignored, as it concerns coverage in the test suite code, and it requires changing the CI to get the coverage there, so I don't think it's worth the effort.

I am questioning the usability of the rule for us. Since most of the project is not covered by tests it will give a lot of warnings.

bobismijnnaam:
> Since most of the project is not covered by tests it will give a lot of warnings.

We'll have to see I guess. We can also turn off the coverage for certain classes/packages, I think, that might help.

PR #659, opened 2021-07-14
Proper error when unfolding abstract predicate

bobismijnnaam:


examples/known-problems/fixed/FunctionProblemFixed1.pvl
```diff
@@ -0,0 +1,13 @@
+//:: cases FunctionProblemFixed1
+//:: tools silicon
+//:: verdict Error
+
```
Vescatur, line 4: In examples/known-problems/FunctionProblem.pvl there was a comment explaining why this fails. It may be since to include it. As "FunctionProblemFixed" is not as descriptive.

src/main/java/vct/col/util/AbstractTypeCheck.java
```diff
@@ -916,6 +918,15 @@ public void visit(OperatorExpression e) {
       case Unfolding: {
         if (!tt[0].isResource()) Fail("Cannot unfold type %s", tt[0]);
         e.setType(tt[1]);
+        MethodInvokation innerMi = getMethodInvokationInsideScale(operatorArgs[0]);
+        if (!(innerMi instanceof MethodInvokation)) {
```
Vescatur, line 922: innerMi is of type MethodInvokation. Why would you check if it is an instanceof MethodInvokation? The only moment !(innerMi instanceof MethodInvokation) is true is when innerMi is null.

Vescatur (changes requested):

bobismijnnaam:
@Vescatur I've processed your comments, can you doublecheck? Then I'll merge it once all the tests pass.

Vescatur:
:)

Vescatur (approved):

PR #661, opened 2021-07-16
Fix simple-ltid.cu

bobismijnnaam:
To fix simple-ltid.cu, typechecking had to be made stricter.

Typechecking is made less lenient, by not forcing arguments of Perm to frac. Specifically, before, given `Perm(x.f, a)`, `a` would always be forced to be typechecked as `frac`, even if it is an `int`. This PR changes it such that `a` is only forced to be a frac if it is a literal (say, 1 or 0). Otherwise, the type is left as is.

By not forcing all possible arguments of perm to frac, some rewrite rules had to be changed to match a frac explicitly, instead of an integer.

Additionally, a rewrite rule is added that rewrites any expression `Perm(f, a)`, where `a` is of type `int`, to `Perm(f, a \ 1)`, such that passing an `int` to a perm works out as expected, instead of causing an error. This was also needed to get simple-litd.cu to work.


src/main/java/vct/col/util/AbstractTypeCheck.java
```diff
@@ -1706,8 +1706,12 @@ private void force_frac(ASTNode arg) {
       Warning("Encountered an integer division ('/') '%s' where a fraction was expected, did you mean a fraction division ('\\') here?", arg);
     }
 
-    if(arg.getType().isPrimitive(PrimitiveSort.Integer)) {
-      arg.setType(new PrimitiveType(PrimitiveSort.Fraction));
+    if (arg instanceof ConstantExpression && arg.getType().isPrimitive(PrimitiveSort.Integer)) {
+      if (arg.equals(0)) {
+        arg.setType(new PrimitiveType(PrimitiveSort.ZFraction));
```
Vescatur, line 1711: We change the type of the argument. Why is that done in TypeCheck? To me "check" signals that it does not have any side effects however since we change the type it does have side effects.

Why do not have to change the value of the argument?

src/main/universal/res/config/simplify_quant_pass1.jspec
```diff
@@ -8,6 +8,7 @@ class simplify_quant_pass1 {
   int i,j,k;
   rational p1,p2;
   int tid;
+  frac f1, f2, f3, f4;
```
Vescatur, line 11: f2, f3 and f4 are never used.

Vescatur (changes requested):

src/main/java/vct/col/util/AbstractTypeCheck.java
bobismijnnaam: Indeed typechecking should be side-effect free, but currently typechecking does have some side effects. I think we will be able to fix that in the new AST, but for now I think we can keep it as is in the old structure.

bobismijnnaam:
After some discussion with @Vescatur we decided to merge the PR.

PR #662, opened 2021-07-16
Reinstate History.pvl

bobismijnnaam:
Changes to get History.pvl working again:
- Ensure RandomizedIf introduces random methods properly (call them with a "this" argument, and generate final methods).
- Also ensure any recursive calls generated by left_merge in CheckProcessAlgebra have a this argument.
- Remove problem-fail from the two cases in History.pvl


Vescatur:
Nice work getting it working again.

Vescatur (changes requested):
Nice work getting it working again. Do you think it is possible to do a null check for create.invokation? If not then you can merge it.

bobismijnnaam:
> Do you think it is possible to do a null check for create.invokation?

Do you mean, inside the definition of create.invokation (so, in ASTFactory.java), check if the first argument is null? I think leaving it null indicates it is a free method call, i.e., not on an object or something but a top-level function call. So null is in fact allowed, it's just not what we want in this case.

bobismijnnaam:
After some discussion with @Vescatur we decided to merge this after the tests pass.

Vescatur:
> Nice work getting it working again. Do you think it is possible to do a null check for create.invokation? If not then you can merge it.

It is not possible to do a null check because null is a valid value.

PR #666, opened 2021-07-16
Issue 663 - Use intellij linter on project

Vescatur:
I wanted to remove all unused imports when I noticed that we can use Intellij to format the code. These are all the changes when we run the code cleanup.

It will probably give some merged conflicts which can be easily fixed. It will also through off the git blame however I think it is worth it for nicely formatted code. If you disagree I could also just remove the unused imports.

I will also look into automatically running the code cleanup.


Vescatur:
The automated code cleanup did not work. I tried again in branch issue-663-2.

PR #667, opened 2021-07-19
Reinstate monotonicBool & monotonicBoolBroken

bobismijnnaam:
This PR removes the rewrite rules concering read (they are now documented here: https://github.com/utwente-fmt/vercors/wiki/Distributing-read-permissions-between-threads). It also refactors the monotonicBool.pvl and monotonicBoolBroken.pvl tests to use fractional permissions for distributing permissions across threads. Unfortunately this makes monotonicBoolBroken.pvl unstable, so this exaple is added to the `problem-partial suite` with `--stop-before-backend`

~The rewrite rules are adjusted to be more limited in how read permissions are rewritten. The result should be similar, except that deep expressions with read will require more rewrite rules now. Luckily, they are pretty rare.~

~The monotonicBool tests are removed from the problemFail suite, and toSeq is replaced by the builtin \values.~

~Since this PR depends on a fix regarding triggers of `\values` in #658, it will keep failing until that PR is merged.~


src/main/universal/res/config/simplify_quant_pass1.jspec
```diff
@@ -432,23 +431,45 @@ class simplify_quant_pass1 {
...
-  // We keep the expression around, because it must trigger an error when it is not well formed, but we don't use it
-  // We're counting on the backend to not optimize it away if it contains an error
-  axiom read_spreads_l {
-    read * i
+  // The next read-permission rules are a bit tricky because the semantics for wildcard/read in viper is a bit limited.
+  // Some discussion:
+  // - https://github.com/viperproject/silicon/issues/539
+  // - https://github.com/viperproject/silicon/issues/569
+  axiom read_spreads_l_i {
+    Perm(g1, read * i)
```
niomaster, line 439: what about negative `i`?

bobismijnnaam:
@niomaster With strict feature checking the test suite prints the following errors:

```  stdout: == Example 1: BindingExpression ==
  stdout: (\exists int i;0 <= i && i < | xs |;xs [ i ])
  stdout: === /home/runner/work/vercors/vercors/examples/parallel/monotonicBool.pvl ===
  stdout:   15   boolean p(int tid);
  stdout:   16 
  stdout:                          [--------------------------------
  stdout:   17   ensures \result == (\exists int i = 0..|xs|; xs[i]);
  stdout:                           --------------------------------]
  stdout:   18   boolean foldOr(seq<boolean> xs) =
  stdout:   19     |xs| > 0 
  stdout: -----------------------------------------
  stdout:   Originating here ^
  stdout: =========================================
  stdout: 
  stdout: 
  stdout: == Example 2: BindingExpression ==
  stdout: (\exists int i;0 <= i && i < N;VCTArray<CT:Ref>.loc(VCTOption<T:VCTArray<CT:Ref>>.getVCTOption1(diz.C_contrib),i).Boolean__item)
  stdout: === /home/runner/work/vercors/vercors/examples/parallel/monotonicBool.pvl ===
  stdout:   56     }
  stdout:   57 
  stdout:                     [----------------------------------
  stdout:   58     assert b == (\exists int i = 0..N; contrib[i]);
  stdout:                      ----------------------------------]
  stdout:   59   }
  stdout:   60 }
  stdout: -----------------------------------------
  stdout:   Originating here ^
  stdout: =========================================```

But the file verifies just fine, so I think viper can do without triggers on existentials. Should I adjust the feature system to allow absence of triggers on `\exists` somehow, or should there be a trigger here? (I don't understand anyway how a trigger would work for an exists)

File: src/main/universal/res/config/simplify_quant_pass1.jspec (hunk 1 above)
Comment (bobismijnnaam) [reply src/main/universal/res/config/simplify_quant_pass1.jspec:439]: That would result in there being no permission for `g1`. I guess we are hiding an error there (since viper wouldn't allow a negative perm anyway, I think?), so that might be a problem.

Comment (bobismijnnaam) [reply src/main/universal/res/config/simplify_quant_pass1.jspec:439]: Viper indeed throws an error upon negative permission, so this rule alters the expected semantics. Not sure if this is worth it, but we do need to address the use case of passing read perms to 2 or more threads, I think.

Comment (niomaster) [reply src/main/universal/res/config/simplify_quant_pass1.jspec:439]: Negative permissions are a falsy well-formedness condition (don't know how else to describe it, it's like injectivity in forall*), so I think that `i >= 0 && i > 0 ==> Perm(g1, read)` is semantics-preserving (though it does not preserve the error)

**Comment by niomaster** [comment]
They are intended as warnings (as the verification completes, yes?) They are part of "permissive features", also known as the list of features I gave up on:

```
vct.col.features.QuantifierWithoutTriggers,
vct.col.features.NestedQuantifiers,
vct.col.features.ExceptionalReturn,
vct.col.features.ContextEverywhere,
```

There are also such warnings on regular quantifiers I think. Triggers are useful in exists when considering the negative of an exists (¬∃φ ↔ ∀¬φ), e.g.:

```
loop_invariant exists i: Int :: {prop(i)}
while(true){}
```

leads to a proof like:

```
assert exists i: Int :: {prop(i)} // pre loop body
assert !(exists i: Int :: {prop(i)}) // post loop body
check-sat
```

roughly equivalent to:

```
assert prop(fresh_i)
assert forall i: Int :: !{prop(i)}
check-sat
```

src/main/universal/res/config/simplify_quant_pass1.jspec
bobismijnnaam: Well, I think I managed to rewrite the example to just use permission a la `Perm(contrib[i], 1 \ N)`, so maybe we no longer need that rule after all. Also, it took quite a bit more steps than before, so I'm suspecting that the previous rewrite rules actually caused some unsoundness... If I can also convert the other example I think we should archive these rewrite rules in the wiki, and do some more thinking about how we want to support this use case in vercors.

bobismijnnaam:
Right, that makes sense. Indeed verification completes, there's another error that causes it to crash, which I interpreted as feature-related, but it's not.

ArmborstL (approved):

PR #669, opened 2021-07-20
Issue 663 2

Vescatur:
This pr removes all unused imports. I first tried to do this automatically however Intellij does not support this for Scala. It is not even possible to get a list of unused imports. I had to go through each file individually. 

I hoped I could integrate the optimization of imports to the workflow but that is unfortunately not possible


bobismijnnaam:
Looks good to me. @niomaster Will this PR cause problems for any of your longer-lived branches?

niomaster:
possibly, but I don't really mind import errors, they are resolved easily of course.

PR #670, opened 2021-07-22
fix sonar test

Vescatur:
This pull request fixes the following warnings in sonarcloud

1) "Shallow clone detected during the analysis. Some files will miss SCM information. This will affect features like auto-assignment of issues. Please configure your build to disable shallow clone."
2) "Could not find ref 'dev' in refs/heads, refs/remotes/upstream or refs/remotes/origin. You may see unexpected issues and changes. Please make sure to fetch this ref before pull request analysis."

1 is because sonar would like to have the git history.
2 is because sonar would like the target branch of the pull request.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -133,8 +133,6 @@ jobs:
     needs: build-test
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
       - uses: actions/setup-java@v1
         with:
           java-version: "11"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -196,6 +194,12 @@ jobs:
     environment: Default
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - run: git checkout dev
```
bobismijnnaam, line 199: Please add a brief comment why these checkouts & fetch depth is necessary, in case the commit information is lost

bobismijnnaam (commented):
1 comment erbij maar verder prima!

PR #671, opened 2021-07-22
Issue 664 - removing unused code

### LINKED ISSUE #664: Remove unused code

Vescatur:
I removed unused code from the folders src, viper, parsers and hre. The folder col is still left. closes #664


Vescatur:
I would like to highlight that the following setting has been removed. "clops.add(sequential_spec.getEnable("sequential specification instead of concurrent"), "sequential") It had no effect in the code.
    "

Vescatur:
The issued highlighted by sonar are violations of the naming conventions. These were already in the code.

col/src/main/java/vct/col/ast/print/AbstractPrinter.java
```diff
@@ -55,18 +55,12 @@ public void nextExpr(){
...
-  
-  private static final Origin missing=new MessageOrigin("unknown location");
-  
+
   public void pre_visit(ASTNode node){
     super.pre_visit(node);
     if (in_expr) {
       expr_level++;
     }
-    Origin o=node.getOrigin();
```
raulmonti, line 66: Is it not possible to get a `none` value here? or how is this managed in case we get one? it will just print `none`?

Vescatur: I am not sure what you mean. The reason I removed this code is because o is never used even if it is assigned. I might have been a programming error however this pull request is not to fix bugs but to remove unused code. I did make an issue for it. https://github.com/utwente-fmt/vercors/issues/675 However I do not think it is worthwhile to investigate it because it is going to be rewriten.

Vescatur: I am not sure what you mean. The reason I removed this code is because o is never used even if it is assigned. I might have been a programming error however this pull request is not to fix bugs but to remove unused code. I did make an issue for it. #675 However I do not think it is worthwhile to investigate it because it is going to be rewriten.

Vescatur:
It looks like I forgot the hre/util folder. I will create a new pull request for that folder. This one is already big enough.

src/main/java/vct/col/util/FeatureScanner.java
```diff
@@ -15,106 +14,42 @@
...
   private EnumSet<Binder> binders_used=EnumSet.noneOf(Binder.class);
-  
-  private HashSet<Class<? extends ASTNode>> nodes=new HashSet<Class<? extends ASTNode>>();
 
   public static FeatureScanner scan(ASTNode node) {
     FeatureScanner fs = new FeatureScanner();
     node.accept(fs);
     return fs;
   }
-
```
raulmonti, line 63: @niomaster we are removing big chunks of this because we don't really used them right now but maybe they are still useful in the short term? Or do you agree we should remove them?

raulmonti (approved):
Looks good, I am surprised of the amount of unreached/unused code we had. I am quite scared of romoving it but I guess that if we need it again then we code it again (better than having a mess around everywhere). I left a comment in one of the files and mentioned Pieter there, but maybe you already know the answer @Vescatur

src/main/java/vct/col/util/FeatureScanner.java
Vescatur: You could say that we are not removing it because it will always be in the git history. However I will discuss with @niomaster whether it is useful to remove all this code.

niomaster: We're not losing particularly complex logic here, I think it's fine to remove this stuff. In fact, we can probably replace FeatureScanner with FeatureRainbow at some point.

PR #673, opened 2021-07-26
Run VeyMont in CI

bobismijnnaam:


examples/veymont-global-programs/TicTacToe.pvl
```diff
@@ -1,6 +1,5 @@
 // -*- tab-width:4 ; indent-tabs-mode:nil -*-
-//:: cases TicTacToeInt TicTacToeMatrix
-//:: tools veymont
```
raulmonti, line 3: why did you remove this line?

raulmonti (approved):
Looks nice to me (loved the sh code :D)

examples/veymont-global-programs/TicTacToe.pvl
bobismijnnaam: The global programs are run with the script in the util folder, so they do not have to be included in vercor's own test suite mechanism (the one triggered by passing the flag ```--test=./examples```). Removing the "tools" directive ensures it's not included in the test suite.

bobismijnnaam: The script is at the bottom of the "files changes" page of this PR, it's called `veymont-2stage-tests.sh`.

raulmonti: yes sorry, I realised that it might be that but did not remove the comment, the thing is that as you mention the script was at the bottom :D

bobismijnnaam:
@raulmonti :) Let's hope the sh code will be short-lived and we can integrate veymont properly soon...

petravandenbos-utwente (approved):
Code looks good.

PR #674, opened 2021-07-30
Speed up PVL Parser

bobismijnnaam:
The PVL parser has a few ambiguities that make it a bit slow. (Parsing the whole test suite takes 32 seconds!) In this PR several of these ambiguities are fixed to speed up parsing of the entire PVL test suite to 6 seconds.

Parsing jspec is also quite slow. This is improved by changing the language for axioms: from `"axiom" "{" langExpr "==" langExpr "}"` to `"axiom" "{" langExpr "}"`. The top-level "==" confuses the expression parsing algorithm because it does not participate in precedence, but it does overlap with the equality token of the binary operator. On my laptop, this approximately halved the time needed to parse all jspec files.

Another reason why the jspec parsing is slow, is because all the special function calls (i.e. "Reducible(A, B)" overlap with function calls in the Java grammar. If I remove these special forms, all jspec files together parse in under 300ms. Unfortunately it's not clear how to fix this and it's likely a somewhat big refactor, so maybe we can work on this in the future.

The only thing left to do in this PR is integrate the debugging code into vercors in such a way that it has a low chance of bit-rotting. E.g. using a certain flag makes vercors print the parsing debugging info, which can be used in one or two of the examples in the CI maybe.


parsers/lib/antlr4/WithPartsCommentedOut_LangPVLParser.g4.bak
```diff
@@ -0,0 +1,335 @@
...
+ | 'join' expr ';'
+ | 'action' tuple block
+ | valStatement
+ | 'if' '(' expr ')' statement elseBlock?
+ | 'barrier' '(' identifier barrierTags? ')' barrierBody
+// | contract 'par' parUnitList
+ | 'vec' '(' iter ')' block
+ | 'invariant' identifier '(' expr ')' block
+ | 'atomic' '(' identifierList ')' block
+// | invariantList 'while' '(' expr ')' statement
```
Vescatur, line 225: On different parts in this file there is commented code.

parsers_sll_ll.csv
```diff
@@ -0,0 +1,345 @@
+name,num_files,bytes,separate_run_millis
```
Vescatur, line 1: What kind of file is this?

Vescatur:
What kind of document is parsers_separate.ods?

parsers/lib/antlr4/WithPartsCommentedOut_LangPVLParser.g4.bak
bobismijnnaam: This is the backup of the file I used to test/profile before I began to actually apply and integrate some performance tweaks I found. So it will be deleted entirely before this PR is merged.

parsers_sll_ll.csv
bobismijnnaam: I think this file contains some performance timings for how long it took to parse each case in SLL/LL mode, and the sizes in bytes.

bobismijnnaam:
It contains the parsing timings of all files/cases, without reusing the parser object, I think.

parsers/lib/antlr4/WithPartsCommentedOut_LangPVLParser.g4.bak
Vescatur: I didn't notice this was a draft before posting my comments.

PR #679, opened 2021-08-19
Issue 678 scalatest

### LINKED ISSUE #678: Converting integration tests to ScalaTest
I would like to explain my vision for converting to ScalaTest and how I am going to make it. Feedback is welcomed. 
-no test code in /src it should all be contained in the test folder. (In case it is necessary I will inject code using dependence injection)
-The test configuration should be moved from the example folder to a scalaTest class. (Otherwise I would need to build some automated conversion between example test to scalaTest. I think that is trying to build a test framework upon a test framework. I also think it adds some benefit of having configuration in code)
-I would like to use the output given to System because it less likely to change compared to the console output.
-It should be possible to specify which passes are run. The execution should be able to stop after verification or before viper. (This is the latest feature Bob made.)
-I am not sure whether I will do the following. Currently the test suite is too slow and big that I am not able to run it locally. It just takes an hour. For local use I would like to speed it up by using a Viper cache. It works as follows, after running a successful test the silver code will saved to a file. In a next run it will check whether the generated silver code equals the saved file. If that is the case it will skip Viper. The cache will not be saved to the repository. It is possible to have a cache in Github action but I have not decided whether I want to.
-The tests should be split among multiple computers for Github actions. I do not know how I will do this. I do not think it is a feature of ScalaTest.

How I am going to approach dependence injection:
Currently we do not use dependence injection. I will add it in two places. In System and in Main. 
-I would like to split Main into multiple classes each doing there own thing. I am currently thinking of one for options, one for passes and one which ties everything together.
-System is static so dependence injection can not be used. What I will do is give System a non static class which does all the work which System does. When testing the class will be replaced by test code.

### LINKED ISSUE #680: Improve github action of ScalaTest
We are currently using a JUnit Test Report action which I do not think as a nice UI. All the error messages are displayed in one long line. It is hard to get an overview of which group of tests are failing. It also does not mention which tests have passed. The action below shows for each tests class which tests have passed.

https://github.com/marketplace/actions/test-reporter

Vescatur:
closes #678 closes #680


examples-not-proccessed/.gitrepo
```diff
@@ -0,0 +1,11 @@
+; DO NOT EDIT (unless you know what you are doing)
```
Vescatur, line 1: Is the example directory still a subrepo?

niomaster: No, this coupling should be removed.

Vescatur:
I have copied some code to new classes. This was to split up the logic of big classes. Sonar gives me warnings that this copied code contains code smells and bugs. I decided to not fix these code smells.

niomaster:
Just posting my notes from the meeting here:
* Perhaps we should tag examples with the original suites from the "::" notation
* The stack size can probably be reduced a bit from 256M (how does stack memory work with multiple threads)
* Investigate if we can leave turbo on
* fix #685

bobismijnnaam:
One more note: the `Runtime / javacOptions` setting can be removed.
   
Ivan's boards where other issues are collected: https://github.com/utwente-fmt/vercors/projects/6, https://github.com/utwente-fmt/vercors/projects/7

niomaster:
This branch is now merged into `ast` via https://github.com/utwente-fmt/vercors/commit/200b85ad5b2aa8c881e4a079454f451fc04dc9ef

Thanks for the contribution and sorry for the delay!

PR #701, opened 2021-10-05
Update Z3 & Boogie

bobismijnnaam:
This PR updates z3 from 4.8.6 to 4.8.12, and boogie, as packaged with carbon, from 21.01 to 21.07. Unfortunately there seem to be some non-deterministic errors for the examples that use z3, so not sure if this PR can be merged as-is. We might have to wait for the next versions to see if it's resolved then, or someone will have to debug it in the meantime.


bobismijnnaam:
The instability of monotonicBool.pvl I cannot reproduce locally. Also, there seems to be an interface problem between viper and z3, as z3 emits a message that viper does not expect. Viper nightly is still packaged with z3 4.8.6, so it seems that the update done in this PR is too early. We should try again with the next viper release (likely 22.01) and update in tandem with them.

PR #713, opened 2021-11-01
Add injectivity/equality axiom to Option<T>

bobismijnnaam:


niomaster (approved):
I believe this is quadratic in the number of occurences of `VCTSome`, but that is a problem we should look at generally. I believe a strategy to manage that is to detected all instances of `(x : option) == (y : option)` in vercors, and replace it with a function `option_eq`, and then trigger on that. That requires some machinery to correctly detect instances of such `==`. For now, let's use this axiom.

PR #736, opened 2022-03-31
COL redesign: Java static imports

bobismijnnaam:
This branch implements some support for static imports in Java.

To do:
- [x] Static field imports
- [x] Fix bug where vercors reuses current class on _all_ static fields
- [x] Static method imports


col/src/main/java/vct/col/resolve/Java.scala
```diff
@@ -203,10 +203,30 @@ case object Java {
...
+      potentialClasses.collect({
+        case RefJavaClass(cls: JavaClass[G]) => cls.getClassField(name)
+      }).flatten match {
+        // If we find only one, or none, then that's good
+        case Seq(field) => Some(field)
+        case Nil => None
+        // Otherwise there is ambiguity: abort
+        // Currently we do not support duplicate imports. E.g. "import static A.X; import static B.*;", given that B
+        // would also define a static X, would technically be allowed.
+        case _ => throw new Exception("Duplicate!") // TODO: Custom error
```
niomaster, line 226: If you want to make a custom error anyway, why not just pick the last appropriate field? I think it's just shadowing in order of imports

PR #737, opened 2022-04-04
Add vim plugin (syntax and filetype detection)

Jesse-Bakker:
These are the syntax files I use when writing PVL in vim. Highlighting is a bit richer than with the textmate grammar.

Example (from `examples/verifythis/2021/TeamBlue/Challenge3.pvl`):

20220404_14h10m11s_grim


niomaster:
Thanks for the contribution!

PR #740, opened 2022-04-28
Minimizer

bobismijnnaam:
This PR is being refactored/rewritten on the [ast-minimize-2](https://github.com/utwente-fmt/vercors/tree/ast-minimize-2) branch.

### Old PR text

This PR implements minimization of the AST, throwing out methods, functions, fields & predicates that the users wants to temporarily ignore.

To do:

- [x] Support for focusing functions
- [x] Abstractify irrelevant functions as well by putting the body in the postcondition
- [x] ~Decide if we want to keep the interface of command line flags. Instead, it could also be an annotation on each indiviual method. Similar to `pure`, but then `focus` and `ignore`. Maybe the functionality in this PR should be merged with the `spec_ignore` concept in VerCors?~ Change syntax to use focus/ignore modifiers. 
- [x] Improve progress output related to minimization. It should be clear to the user whenever a method is not checked for some reason, as this diverges from VerCors' normal behaviour, that is, verify everything. Preferably with readable names, e.g. in Java the fully qualified name should be used. 
- [x] Should the `focus`/`ignore` attributes be refactored as first-class properties (like `pure` and `inline` of instancemethod, instancefunction)? This would resolve the tension with origins, and allow to move the origin trickery out of JavaToCol
- [ ] Add tests for java constructors, top level functions/procedures
- [x] PVL support
- [ ] Silver support
- [ ] C support


niomaster:
Shall we close this in favor of ast-minimize-2?

bobismijnnaam:
Sure, though I would like the old branch to stick around for a bit, if that's okay.

PR #792, opened 2022-10-03
Unsound & incomplete float support

bobismijnnaam:
Implements support for floats in C, PVL, and Java.

- Only strict typechecking rules: safe conversion/widening is allowed (float -> double), but not back (which is allowed in C, but not in Java).
- Unsafe conversion can be done with explicit casts (e.g. `(float) 0.5` in Java/C, `(0.5).toFloat32` in PVL). This removes all knowledge of the value before conversion.
- Floats are unsoundly (!) encoded into rationals
- Not sure what happens if you put NaN somewhere


src/test/scala/vct/test/integration/helper/VercorsSpec.scala
```diff
@@ -139,7 +140,7 @@ abstract class VercorsSpec extends AnyFlatSpec {
   }
 
   class DescPhrase(val verdict: Verdict, val backends: Seq[Backend], val desc: String) {
-    def pvl(data: String): Unit = {
+    def pvl: String => Unit = { data =>
```
niomaster, line 143: if this doesn't work, please remove the change

src/test/scala/vct/test/integration/helper/VercorsSpec.scala
```diff
@@ -168,5 +169,6 @@ abstract class VercorsSpec extends AnyFlatSpec {
 
   val silicon: Seq[Backend] = Seq(types.Backend.Silicon)
   val carbon: Seq[Backend] = Seq(types.Backend.Carbon)
+  val someBackend: Seq[Backend] = Seq(types.Backend.Silicon)
```
niomaster, line 172: what's the reasoning for this?

niomaster (changes requested):

PR #810, opened 2022-10-10
Better support for static imports, fields, and methods

bobismijnnaam:
This PR refactors Resolution & ResolveTypes so it's possible to distinguish between local names and imported names and types. This also allows static method calls like `System.sqrt` to work.

It was a problem that vercors could not derive that `System.out` is non-null, which is guaranteed by the JVM. ~This information can be encoded on a case by case basis (some sort of static field annotation) or by generalizing this use case (some sort of static class (for System.out) and/or immutable class (for java.lang.String) invariant).~ This problem was solved ad-hoc by assuming this information where it is needed.

### DIFF
--- a/col/src/main/java/vct/col/resolve/Resolve.scala
+++ b/col/src/main/java/vct/col/resolve/Resolve.scala
@@ -32,7 +32,23 @@ case object ResolveTypes {
     case Program(decls, _) =>
       ctx.copy(stack=decls.flatMap(Referrable.from) +: ctx.stack)
     case ns: JavaNamespace[G] =>
-      ctx.copy(stack=ns.declarations.flatMap(Referrable.from) +: ctx.stack, namespace=Some(ns))
+      // Static imports need to be imported at this stage, because they influence how names are resolved.
+      // E.g.: in the expressio f.g, f is either a 1) variable, 2) parameter or 3) field. If none of those, it must be a
+      // 4) statically imported field or typename, or 5) a non-static imported typename. If it's not that, it's a package name.
+      // ctx.stack needs to be modified for this, and hence this importing is done in enterContext instead of in resolveOne.
+      val staticallyImportedTypes: Seq[Referrable[G]] = ns.imports.collect {
+        case imp @ JavaImport(/* static = */ true, JavaName(fullyQualifiedTypeName :+ staticMember), /* star = */ false) =>
+          val staticType = Java.findJavaTypeName(fullyQualifiedTypeName, ctx)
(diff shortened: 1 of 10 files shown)


col/src/main/java/vct/col/resolve/Resolve.scala
```diff
@@ -82,18 +99,34 @@ case object ResolveTypes {
...
+//
+//    case imp @ JavaImport(/* static = */ true, name, /* star = */ true) =>
+//      Java.findJavaTypeName(name.names, ctx)
+//        .getOrElse(throw NoSuchNameError("class", name.names.mkString("."), imp))
+//    case imp@JavaImport(/* static = */ false, name, /* star = */ false) =>
+//      Java.findJavaTypeName(name.names, ctx)
+//        .getOrElse(throw NoSuchNameError("class", name.names.mkString("."), imp))
+//    case imp@JavaImport(/* static = */ false, pkg, /* star = */ true) =>
+//      // TODO (RR): Pulls in everything, even though we prefer on demand...?
+//      Java.findJavaTypeNamesInPackage(pkg.names, ctx)
```
niomaster, line 129: ?

col/src/main/java/vct/col/resolve/lang/Java.scala
```diff
@@ -97,6 +95,8 @@ case object Java {
         ((t : JavaNamedType[_]).unsafeTransmuteGeneration[JavaNamedType, G] : JavaNamedType[G]).ref = Some(RefJavaClass(colClass))
       }
 
+      println(s"[warning] No specification was found for class ${potentialFQName.mkString(".")}, so a shim was loaded from the JRE.")
```
niomaster, line 98: oh, this should be a logging line

src/test/scala/vct/test/integration/helper/VercorsSpec.scala
```diff
@@ -38,6 +38,18 @@ abstract class VercorsSpec extends AnyFlatSpec {
...
+//    val fTrait = new AnyFlatSpecLike{}.getClass.getDeclaredFields.toList.find(_.getName.contains("engine")).get
+//    val f = this.getClass.getSuperclass.getDeclaredField(fTrait.getName)
+//    f.setAccessible(true)
+//    val engine = f.get(this)
+//    val engineClass = engine.getClass
+//    val method = engineClass.getSuperclass.getDeclaredMethods.toList.find(_.getName.contains("registerTest")).get
+//    method.setAccessible(true)
+//    method.invoke(engine,
+//      null
+//    )
```
niomaster, line 51: ?

niomaster (changes requested):

PR #821, opened 2022-10-19
Derive classpath from sbt file

niomaster:
Can someone on windows please test this? @ArmborstL @PBHTasche


niomaster:
it appears that fullClasspath is not populated until inspected, and fullClasspath does depend on compilation, which is very annoying...

PR #854, opened 2022-10-31
CPP support

OmerSakar:
This PR implements support for CPP.


niomaster:
@OmerSakar Any chance you could fix up this PR so it compiles and doesn't break other VerCors components, so we can get this bit merged? That way you benefit from refactorings in `dev` in the meantime when you continue.

pieter-bos:
@OmerSakar @Ellen-Wittingen From what I understand this branch is no longer needed for further C++ support work. If not, please speak up now or forever hold your peace :)

PR #868, opened 2022-11-11
Allow assume to flatten side effects to before the statement

bobismijnnaam:


src/test/scala/vct/test/integration/examples/TechnicalSpec.scala
```diff
@@ -348,4 +348,15 @@ class TechnicalSpec extends VercorsSpec {
...
+  vercors should verify using anyBackend in "assigning results from impure method calls to final fields should preserve information" java
+    """
+      class C {
+          final Integer x;
+          //@ ensures this.x != null && this.x.intValue() == 0;
+          C() {
+              this.x = new Integer(0);
+          }
+      }
+    """.stripMargin
```
niomaster, line 361: There is no margin to strip

niomaster (dismissed):

PR #878, opened 2022-11-24
Serialize

niomaster:


niomaster:
Even with splitting and caching disabled, the verification is still significantly slower than in dev, why?

Issue #820, opened 2022-10-18
Passing a sized array as a parameter to a C function is incorrectly interpreted by vercors


Naum-Tomov:
The following code:
```
void zero(int M, int matrix[M]){
  for(int i=0; i<M; i++)
  /*@
    requires matrix != NULL;
    context Perm(matrix[i], write);
    ensures matrix[i] == 0;
  @*/
  {
      matrix[i] = 0;
  }
}
```
throws the following exception: `[ERROR] Unrecoverable error: Trying to subscript (matrix[i]) a non subscriptable variable with type int`
This is thrown via AmbiguousSubscript, I suspect, because VerCors assumes matrix is of type int.
Conversely, if we modify the code such that matrix is not a function parameter, e.g.
```
void zero(){
  int M;
  int matrix[M];
  for(int i=0; i<M; i++)
  /*@
    requires matrix != NULL;
    context Perm(matrix[i], write);
    ensures matrix[i] == 0;
  @*/
  {
      matrix[i] = 0;
  }
}
```
The verification completes successfully.
Furthermore, the exception message itself does not point to which part of the code is causing the error which might also be an issue in and of itself.

Issue #791, opened 2022-09-29
True matrices


niomaster:
We need real multi-dimensional array support, since C requires it, and it would be nice to have in PVL. Let's also deal with the linearized matrix indexing problem at the same time (i.e. `ar[i*N+j]` where `ar.length == N*M, i < M, j < N`)

dzagieboylo:
I've been looking into this as it is needed by lots of CUDA programs (e.g., tiling via statically sized `__shared__` memory is very common).

There were some bugs (e.g., the type conversion of `CTArray` always flattened multi-dimensional arrays and didn't accurately track the inner pointer type) that I've fixed in my fork; however, I'm still running into a *major issue*.

The declaration `int arr[4]` gets translated to:
```
arr = new pointer<int>[4];
```
Which lets us establish facts about the (one-D) array's locations and permissions.

However, the declaration `int arr[4][2]` also just gets translated to a one liner:

```
arr = new pointer<pointer<int>>[4];
```
Which doesn't capture the fact that the inner arrays are also automatically allocated and write-permissible in the C semantics when provided static bounds.

I've added a translation which automatically generates this initialization code in `LangCToCol`  but it's actually ALSO difficult to generate the information for VerCors that all of the pointers are unique and correctly allocated; this makes using MD arrays impractical.

------

Is it possible to have better backend support for MD arrays that automatically establishes the write permissions for each location? I'm also happy to try to implement it, if given some pointers where to look.

superaxander:
Yes we need support proper support for (multi-dimensional) arrays still. Treating them as pointers is incorrect since for example the following program verifies:

```c
int main(void) {
    int a = 10;
    int b[2][3];

    b[0] = &a;

    return 0;
}
```

but it doesn't compile since b[0] is of type int[] which cannot be assigned a pointer (in fact it is not assignable at all)

Another example:

```
struct A {
    int a[10];
};

int main(void) {
    struct A b;
    int c[10];
    b.a = c;

    return 0;
}
```

Yeah arrays are not pointers and we should not treat them as such. It's one of the things on my to do list. 

To have some rudimentary support for multi-dimensional arrays it would be fine for now to do something like the translation you mentioned and then a user could use something like `\matrix` but for pointers which would automatically include all the necessary quantifiers to show that all the elements are unique.

superaxander:
I'm having a closer look at a proper encoding for arrays right now.

PR #893, opened 2022-12-13
Sized array parameter

Naum-Tomov:
Resolved the issue #820 by adding a CTArray type that contains the size of the array as a parameter. However, due to lack of true matrices, a lot of the examples associated still do not pass, and have been moved to #791.


niomaster (dismissed):
Please have a look at the failing test before merging

Issue #847, opened 2022-10-27
`CoercingRewriter` is not complete yet


niomaster:
All other node types (that don't have a `coerce` yet) should get one.

PR #900, opened 2022-12-16
Finish coercing rewriter

Naum-Tomov:
closes issue #847


col/src/main/java/vct/col/typerules/CoercingRewriter.scala
```diff
@@ -250,6 +253,109 @@ abstract class CoercingRewriter[Pre <: Generation]() extends Rewriter[Pre] with
   override def dispatch(region: ParRegion[Pre]): ParRegion[Post] =
     postCoerce(coerce(preCoerce(region)))
 
+
+
+  def preCoerce(node: Verification[Pre]): Verification[Pre] = node
+  def postCoerce(node: Verification[Pre]): Verification[Post] = rewriteDefault(node)
+  override def dispatch(node: Verification[Pre]): Verification[Rewritten[Pre]] = postCoerce(coerce(preCoerce(node)))
```
niomaster, line 260: Could you please try making all the `dispatch`es `final`? This prevents the mistake of overriding it out of habit (and will break compilation of most `Import*` passes: we should replace `dispatch` with `postCoerce` in those passes)

col/src/main/java/vct/col/typerules/CoercingRewriter.scala
```diff
@@ -1301,4 +1407,260 @@ abstract class CoercingRewriter[Pre <: Generation]() extends Rewriter[Pre] with
...
+
+  def coerce(node: DecreasesClause[Pre]): DecreasesClause[Pre] = {
+    implicit val o: Origin = node.o
+    node match {
+      case DecreasesClauseAssume() =>
+        DecreasesClauseAssume()
+      case DecreasesClauseNoRecursion() =>
+        DecreasesClauseNoRecursion()
+      case DecreasesClauseTuple(exprs) =>
+        DecreasesClauseTuple(exprs)
```
niomaster, line 1497: I changed my mind about this: since we currently only support integers, lets coerce the expressions to `int`.

niomaster (changes requested):

col/src/main/java/vct/col/typerules/CoercingRewriter.scala
Naum-Tomov: Which dispatch should be replaced, the one in CoercingRewriter, or the one in the Import* class?

niomaster: The ones in the Import classes

niomaster (approved):

PR #904, opened 2022-12-20
Enable adts for java

bobismijnnaam:
Enable ADTs in Java.


col/src/main/java/vct/col/ast/Node.scala
```diff
@@ -75,7 +75,7 @@ final case class Program[G](declarations: Seq[GlobalDeclaration[G]])(val blame:
 sealed trait Type[G] extends NodeFamily[G] with TypeImpl[G]
 
 object TNotAValue {
-  def unappply[G](t: TNotAValue[G]): Some[Referrable[G]] = Some(t.decl.get)
+  def unapply[G](t: TNotAValue[G]): Some[Referrable[G]] = Some(t.decl.get)
```
niomaster, line 78: i hApPeN To lIkE My uNaPpLy wItH An eXtRa p

niomaster (approved):

PR #911, opened 2022-12-21
fix #838: allow nearby-import for java, generalize classpath notion

### LINKED ISSUE #838: Reconsider nearby-import for java
Previously we had a feature where VerCors would search the directory of the verified file(s) (taking into account the `package`) for imported classes that are used in the file, so we can import a contract outline.

niomaster:


niomaster:
This PR does not make any methods abstract, nor does it consider whether imported entities are transparent or not (e.g. functions, and pure/inline methods). IMO this PR is an improvement regardless of that caveat, since earlier we simply did not resolve files near to the source file.

PR #918, opened 2022-12-28
making example to work in javabip

lsafina:
Hi Bob! 
I've changed the artefact example so it can be run in JavaBIP, can you please have a look if it still works for Vercors?


examples/concepts/javabip/casinoBroken/Casino.java
```diff
@@ -71,8 +83,8 @@ public void createGame(@Data(name = OPERATOR) Integer sender) {
 
     // Operator receives a bet
     @Transition(name = RECEIVE_BET, source = GAME_AVAILABLE, target = BET_PLACED,
-            guard = "IS_NOT_OPERATOR && ALLOWABLE_BET",
-            pre = "0 <= bet && guess != null")
+            guard = "IS_NOT_OPERATOR & ALLOWABLE_BET",
+            requires = "0 <= bet && guess != null")
```
bobismijnnaam, line 87: Good point. I need to update the names still.

examples/concepts/javabip/casinoBroken/Casino.java
```diff
@@ -84,7 +96,7 @@ public void receiveBet(@Data(name = PLAYER) Integer sender,
     }
 
     @Transition(name = CASINO_WIN, source = BET_PLACED, target = IDLE, guard = "IS_OPERATOR & !GUESSED")
-    public void casinoWin(@Data(name = OPERATOR) Integer sender) {
+    public void casinoWin(/*@Data(name = OPERATOR) Integer sender*/) {
```
bobismijnnaam, line 99: Is this an actual technically needed change, or are the arguments just not necessary here? (As they are not used I would agree they seem useless)

examples/concepts/javabip/casinoBroken/Casino.java
```diff
@@ -95,7 +107,7 @@ public void casinoWin(@Data(name = OPERATOR) Integer sender) {
     }
 
     @Transition(name = PLAYER_WIN, source = BET_PLACED, target = IDLE, guard = "IS_OPERATOR & GUESSED & IS_PLAYER")
-    public void playerWin(@Data(name = PLAYER) Integer player, @Data(name = OPERATOR) Integer operator) {
+    public void playerWin(/*@Data(name = PLAYER) Integer player, @Data(name = OPERATOR) Integer operator*/) {
```
bobismijnnaam, line 110: Same here, technicality or just cleanup?

examples/concepts/javabip/casinoBroken/Constants.java
```diff
@@ -47,7 +47,6 @@ public class Constants {
     public static final String OUTGOING_FUNDS = "OUTGOING_FUNDS";
     public static final String INCOMING_FUNDS = "INCOMING_FUNDS";
     public static final String AVAILABLE_FUNDS = "AVAILABLE_FUNDS";
-    public static final String CASINO = "CASINO";
```
bobismijnnaam, line 50: There is already a `CASINO` constant above, right? Not sure why I added this one...

examples/concepts/javabip/casinoBroken/Constants.java
```diff
@@ -57,7 +56,6 @@ public class Constants {
     public static final String GUESSED = "GUESSED";
     public static final String ENOUGH_FUNDS = "ENOUGH_FUNDS";
     public static final String ALLOWABLE_BET = "ALLOWABLE_BET";
-    public static final String SAFE_GAME_STEP = "SAFE_GAME_STEP";
```
bobismijnnaam, line 60: Is this deleted because it is unused or because of a technicality?

I wanted to keep the diff between the broken & adjusted version as small as possible, so personally I would like to keep "unused but safe" definitions in there if possible

examples/concepts/javabip/casinoBroken/Constants.java
bobismijnnaam: Indeed works fine without the extra public static final

examples/concepts/javabip/casinoBroken/Casino.java
bobismijnnaam: Applied the renaming, finally

examples/concepts/javabip/casinoBroken/Casino.java
bobismijnnaam: Ah, I guess the IDE gives a warning or compile error if arguments are removed. Best to remove them, then.

bobismijnnaam: No, wait, it is used: by the IS_OPERATOR guard, right? Then we should not remove it.

examples/concepts/javabip/casinoBroken/Casino.java
bobismijnnaam: Again shouldn't be removed because of the guard?

niomaster:
If I've understood correctly this pull request is now addressed elsewhere, and can be closed. Please feel free to re-open it if that is not correct.

PR #936, opened 2023-02-01
Java static support

### LINKED ISSUE #781: New resolution system does not yet understand static field access
The following code:

```
class C {
    void m() {
        //@ assert System.out == null;
    }
}
```

Has output:

```
[INFO] Starting verification
======================================
At /home/bobe/Downloads/test.java:3:20:
--------------------------------------
    1  class C {
    2      void m() {
                         [------
    3          //@ assert System.out == null;
                          ------]
    4      }
    5  }
--------------------------------------
Could not find local named System.
======================================
```

I guess we should also think about how permissions for static fields are managed...

### LINKED ISSUE #912: Reference to package-qualified class not found
In Java, you can reference a class from `java.util` either by first importing it and then using its name, or by referring to it with the full package-qualified name without an import.
VerCors apparently can only resolve the first, not the second.
I.e. this works:
```java
import java.util.Arrays;

class C {
  void m() {
    Arrays.fill(new Object[3], 0, 1, null);
  }
}
```
while this fails:
```java
class C {
  void m() {
    java.util.Arrays.fill(new Object[3], 0, 1, null);
  }
}
```
with VerCors saying _"Could not find method named fill."_
Javac can deal with both.

<details>
<summary> Arrays.java</summary>
Note: I added the following file in `src/main/universal/res/jdk/java/util`:

```java
package java.util;

public class Arrays {
  public static void fill(Object[] arr, int start, int end, Object value) {
  }
}
```
</details>

Naum-Tomov:
We currently support statics but if we import a class with an error we forward the error for the current verification, which we ultimately do not desire.
We have static imports, the possibility of importing static fields and methods and using them in other classes in the same package.
A new test file TechnicalStaticSpec included that tests the features.
Missing final const expr.
Needs an optimization for System and other java static classes (like java.util) that load too many classes and make verification very slow.
Closes #912. (at least it gets past the resolution stage safely, needs very long to actually verify on my machine, but I would assume it does)
Closes #781 (same as previous issue, gets past resolution stage, cannot know if it actually verifies, without the needed optimization)


Naum-Tomov:
These last commits implement the final constant expressions. It is now possible to assert things that involve the value of a final integer field.

col/src/main/java/vct/col/resolve/lang/Java.scala
```diff
@@ -92,6 +112,11 @@ case object Java extends LazyLogging {
     }
   }
 
+  def findRuntimeJavaTypesInPackage[G](pkg: Seq[String], ctx: TypeResolutionContext[G]): Seq[JavaClassOrInterface[G]] = {
+    // TODO...
+    Nil
+  }
+
```
niomaster, line 119: this is obsolete I think

niomaster (dismissed):

PR #949, opened 2023-02-10
Silicon progress tracing

bobismijnnaam:
This PR adds rudimentary silicon progress tracing in vercors by pretty printing all origins in the current trace maintained in symbexlogger. It uses the jAnsi library to control the terminal cursor, clear the screen, etc.

This proof of concept works but more work is needed to make it just as friendly as the current progress printer: there should be a flag to turn it off, it shouldn't be on when stdout is not a tty, it should not use placeholder strings like "XXX", etc. Not sure if it is useful enough to warrant the investment, but it certainly looks cool.

### DIFF
--- a/rewrite/src/main/java/vct/col/rewrite/SimplifyNestedQuantifiers.scala
+++ b/rewrite/src/main/java/vct/col/rewrite/SimplifyNestedQuantifiers.scala
@@ -518,6 +518,8 @@ case class SimplifyNestedQuantifiers[Pre <: Generation]() extends Rewriter[Pre] 
 
   case class Pointer[G](index: Expr[G], subnodes: Seq[Node[G]], array: Expr[G]) extends Subscript[G]
 
+  case class Sequence[G](index: Expr[G], subnodes: Seq[Node[G]], array: Expr[G]) extends Subscript[G]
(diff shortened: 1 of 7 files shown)


niomaster:
I think it looks quite nice now, please let me know soon if you would like to keep the work in this branch around

rewrite/src/main/java/vct/col/rewrite/SimplifyNestedQuantifiers.scala
```diff
@@ -518,6 +518,8 @@ case class SimplifyNestedQuantifiers[Pre <: Generation]() extends Rewriter[Pre]
 
   case class Pointer[G](index: Expr[G], subnodes: Seq[Node[G]], array: Expr[G]) extends Subscript[G]
 
+  case class Sequence[G](index: Expr[G], subnodes: Seq[Node[G]], array: Expr[G]) extends Subscript[G]
+
```
bobismijnnaam, line 522: @niomaster This is the only change of interest, and it's not actually related to the intention of the silicon-progress-tracing branch. I think it's also in #950. So as far as I'm concerned this branch can be deleted.

PR #950, opened 2023-02-10
Small debug fixes

bobismijnnaam:
Some debug fixes I needed:

- Make parallelblockencoder origins a bit more precise
- Enable Lars's rewriting system for sequences as well (it only worked on arrays before).

If the test suite still runs okay with lars's code enabled for sequences we could consider keeping that in, but not sure if it's worth the risk. VerCors not touching any quantifiers randomly is also a nice property...


niomaster (approved):
modulo the test suite I have no objections to merge this.

PR #953, opened 2023-02-14
Negative array sizes

### LINKED ISSUE #416: Negative array sizes
Currrently VerCors allows negative array sizes, both in the Java and PVL frontend. Java throws an exception in this case, so at least in the java frontend it should be an error (unless we want to allow it if the function declares the proper `throws` clause - which I am currently not planning to allow, unless someone objects). For PVL I think it would be good to forbid it as well. It can probably be encoded straightforwardly by emitting an assert before array creation, asserting that it should be `>= 0`. Any objections? (Unless this is already implemented - then it must be currently broken)

For reference, I used the following program:

```
public class ArrayBounds {
    public void foo() {
        int[] arr = new int[-1];
    }
}
```

Both `ArrayBounds.java` and `ArrayBounds.java --lang=pvl` verify with the current dev branch.

Naum-Tomov:
closes #416


niomaster:
@sakehl - could you have a look at this?

Context: We want to make it a verification error to declare an array (or pointer memory block) of negative size, see also #416

https://github.com/utwente-fmt/vercors/blob/7b7dbeb21bcecdcc86bed97846bd0994d5a2fa39/rewrite/src/main/java/vct/col/rewrite/lang/LangCToCol.scala#L360-L366

If I understand correctly: the size of a dynamic shared memory block is declared as a fresh argument. Some thoughts:
* Do we account that in such a way that we can still support kernel invocations?
* Should we just `require` that these variables are zero? I'm worried we might trigger the PanicBlame otherwise

sakehl:
@niomaster 

Sorry I was out of the office for a while, thus my delayed response.

> * Do we account that in such a way that we can still support kernel invocations?
In CUDA the size of the dynamic shared memory is an extra argument in the kernel invocation. Thus we should be able to support that. (See [here](https://developer.nvidia.com/blog/using-shared-memory-cuda-cc)

In OpenCL this is different but I'd think it should be possible.

> * Should we just require that these variables are zero? I'm worried we might trigger the PanicBlame otherwise
That sounds like a good idea. I think you mean "bigger or equal to zero" though right?

niomaster:
Thanks! All sounds good. non-negative indeed. I opened #977 to track this.

PR #969, opened 2023-02-24
[VCLLVM] setup LLVM as supported VerCors language

Drevanoorschot:
A lot is still missing such as passing an actual file to VCLLVM and parsing the buffer back to a `ParseResult`, but at least it compiles :).


niomaster:
The CI doesn't run for forks (rather than branches in this repository), I'll look into doing something about that. For now this doesn't seem to affect the test suite.

PR #980, opened 2023-03-07
[VCLLVM] LLVMFunctionContract & LLVMFunctionDefinition

Drevanoorschot:
This PR creates two new node types in the COL AST. Their rewrite rules still need to be implemented, i.e. rules to convert `LLVMFunctionContract` => `ApplicableContract` and `LLVMFunctionDefinition` => `Procedure`.

I followed the the [guide](https://github.com/utwente-fmt/vercors/wiki/Nodes) to the best of my ability. It compiles but throws more warnings (specifically about coercion) than before I added the nodes so I wouldn't be surprised if I got something wrong. :see_no_evil:

### DIFF
--- a/col/src/main/java/vct/col/ast/Node.scala
+++ b/col/src/main/java/vct/col/ast/Node.scala
@@ -908,14 +908,14 @@ final case class BipInternal[G]()(implicit val o: Origin = DiagnosticOrigin) ext
 final case class BipPortSynchronization[G](ports: Seq[Ref[G, BipPort[G]]], wires: Seq[BipGlueDataWire[G]])(val blame: Blame[BipSynchronizationFailure])(implicit val o: Origin) extends GlobalDeclaration[G] with BipPortSynchronizationImpl[G]
 final case class BipTransitionSynchronization[G](transitions: Seq[Ref[G, BipTransition[G]]], wires: Seq[BipGlueDataWire[G]])(val blame: Blame[BipSynchronizationFailure])(implicit val o: Origin) extends GlobalDeclaration[G] with BipTransitionSynchronizationImpl[G]
 
-final class LLVMFunctionContract[G](val value:String, val references:Seq[(String, Ref[G, Declaration[G]])])(implicit val o: Origin) extends NodeFamily[G] with LLVMFunctionContractImpl[G]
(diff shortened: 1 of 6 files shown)


col/src/main/java/vct/col/ast/Node.scala
```diff
@@ -908,6 +908,16 @@ final case class BipInternal[G]()(implicit val o: Origin = DiagnosticOrigin) ext
 final case class BipPortSynchronization[G](ports: Seq[Ref[G, BipPort[G]]], wires: Seq[BipGlueDataWire[G]])(val blame: Blame[BipSynchronizationFailure])(implicit val o: Origin) extends GlobalDeclaration[G] with BipPortSynchronizationImpl[G]
 final case class BipTransitionSynchronization[G](transitions: Seq[Ref[G, BipTransition[G]]], wires: Seq[BipGlueDataWire[G]])(val blame: Blame[BipSynchronizationFailure])(implicit val o: Origin) extends GlobalDeclaration[G] with BipTransitionSynchronizationImpl[G]
 
+final class LLVMFunctionContract[G](val value:String, val references:Set[(String, Ref[G, Declaration[G]])])(implicit val o: Origin) extends NodeFamily[G] with LLVMFunctionContractImpl[G]
```
niomaster, line 911: I recommend against using `Set`s in the AST, since it's more prone to introduce non-determinism in VerCors. For now please use a `Seq`, and we'll likely support `ListSet` in the future.

Drevanoorschot:
The latest [commit](https://github.com/utwente-fmt/vercors/pull/980/commits/da8108aef61b617ad6b94d4f85542cf7ed043b26) converts the aforementioned `Set` to a `Seq`

Drevanoorschot:
On a side note, would it be a problem code style wise to change any node type currently named `LLVM*` to `llvm*`? The C++ code generator of protobuf converts camelcased message types to snake case resulting in these... interesting function names:
```C++
pProgram->add_declarations()->mutable_l_l_v_m_function_definition();
```
Changing the COL node type definition of LLVM nodes in VerCors to lowercase would resolve the issue, but i can see how it might be an issue codestyle wise.

niomaster:
No problem, starting with `Llvm` is fine. This is actually probably due to our proto generation, see also e.g. `a_d_t_function_invocation`.

PR #981, opened 2023-03-08
Systemc2vercors

PBHTasche:
VeSUV integration into VerCors.


parsers/src/main/java/vct/parsers/transform/systemctocol/exceptions/UnsupportedException.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public String text() {
+        return "Unsupported object: " + text;
+    }
+
+    @Override
+    public String code() {
+        return sc_obj.toString();
```
niomaster, line 26: codes should be constants, they're mostly internal to the testing framework

niomaster (changes requested):

niomaster (approved):

Issue #989, opened 2023-03-20
\unfold in precondition causes information loss


JLedelay:
The following input: 
```java
public class C {
  final Object[] arr;

  /*@
    resource P() =
      arr != null ** 
      Perm(arr[*], write);
  @*/

  /*@
    requires arr != null ** P();
    requires (\forall int i = 0 .. arr.length; \unfolding P() \in {:arr[i]:} != elem);
  @*/
  public void addP(Object elem) {
    //@ assert (\forall int i = 0 .. arr.length; \unfolding P() \in {:arr[i]:} != elem);
  }

  /*@
    requires arr != null ** P();
    requires (\forall int i = 0 .. arr.length; \unfolding P() \in {:arr[i]:} != elem);
  @*/
  public void addP2(Object elem) {
    //@ unfold P();
    //@ assert (\forall int i = 0 .. arr.length; {:arr[i]:} != elem);
  }

  /*@
    requires arr != null ** Perm(arr[*], write);
    requires (\forall int i = 0 .. arr.length; {:arr[i]:} != elem);
  @*/
  public void add(Object elem) {
    //@ assert (\forall int i = 0 .. arr.length; {:arr[i]:} != elem);
  }
}
```

gives the following output:
```
======================================
 At C.java:15:7:
--------------------------------------
   13   @*/
   14   public void addP(Object elem) {
                   [--------------------------------------------------------------------------------
   15           //@ assert (\forall int i = 0 .. arr.length; \unfolding P() \in {:arr[i]:} != elem);
                    --------------------------------------------------------------------------------]
   16   }
   17
--------------------------------------
[1/2] Assertion may not hold, since ...
--------------------------------------
 At C.java:15:14:
--------------------------------------
   13   @*/
   14   public void addP(Object elem) {
                          [------------------------------------------------------------------------
   15           //@ assert (\forall int i = 0 .. arr.length; \unfolding P() \in {:arr[i]:} != elem);
                           ------------------------------------------------------------------------]
   16   }
   17
--------------------------------------
[2/2] ... this expression may be false (https://utwente.nl/vercors#assertFailed:false)
======================================
======================================
 At C.java:24:7:
--------------------------------------
   22   public void addP2(Object elem) {
   23           //@ unfold P();
                   [-------------------------------------------------------------
   24           //@ assert (\forall int i = 0 .. arr.length; {:arr[i]:} != elem);
                    -------------------------------------------------------------]
   25   }
   26
--------------------------------------
[...]

PR #1005, opened 2023-04-06
fix inline triggers with \unfolding

Naum-Tomov:
change Transformation.scala to _also_ extract inline triggers before importing arrays, so that ImportArray.rewriteTopLevelArraySubscriptsInTrigger may do away with references of array subscripts in inline triggers


Naum-Tomov:
This closes #989

PR #1037, opened 2023-06-06
Fixes for examples

ArmborstL:
A few minor changes needed to verify some examples I used recently (e.g. the ArrayList program from Joost).

### DIFF
--- a/res/universal/res/jdk/java/lang/Integer.java
+++ b/res/universal/res/jdk/java/lang/Integer.java
@@ -1,6 +1,9 @@
 package java.lang;
 
 class Integer {
+    static int MIN_VALUE;
+    static int MAX_VALUE;
+    
     //@ ensures intValue() == x;
     public Integer(int x) {
         //@ assume false;
--- a/res/universal/res/jdk/java/lang/Integer.java
+++ b/res/universal/res/jdk/java/lang/Integer.java
@@ -1,8 +1,8 @@
 package java.lang;
 
 class Integer {
-    static int MIN_VALUE;
-    static int MAX_VALUE;
+    final static int MIN_VALUE;
+    final static int MAX_VALUE;
     
     //@ ensures intValue() == x;
     public Integer(int x) {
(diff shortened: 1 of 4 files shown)


src/col/vct/col/resolve/lang/Spec.scala
```diff
@@ -15,7 +15,7 @@ case object Spec {
     case RefInstanceMethod(decl) => decl.contract
     case RefCLocalDeclaration(decls, _) => decls.decl.contract
     case RefCGlobalDeclaration(decls, _) => decls.decl.contract
-//    case RefCFunctionDefinition(decl) =>
+    case RefCFunctionDefinition(decl) => decl.contract
```
niomaster, line 18: I think this is commented out because the contract of a function definition may also be the contract of the pre-declaration, if it is empty at the definition.

res/universal/res/jdk/java/lang/Integer.java
```diff
@@ -1,6 +1,9 @@
 package java.lang;
 
 class Integer {
+    static int MIN_VALUE;
+    static int MAX_VALUE;
```
niomaster, line 5: `final`?

src/col/vct/col/resolve/lang/Spec.scala
ArmborstL: Okay, so it should be something like `decl.contract ? decl.contract : pre_decl.contract`? What is the pre-declaration / where does it come from? nothing in `CFunctionDefinition` immediately looks fitting...

res/universal/res/jdk/java/lang/Integer.java
ArmborstL: yeah, makes sense. I'll update it.

src/col/vct/col/resolve/lang/Spec.scala
niomaster: haha hence why it was not immediately solved :) I don't think it's directly available here. I'll have a look on what to do

niomaster: Actually there seems to be no logic at all connecting forward declarations and definitions, so I'm ok leaving it like this for now.

PR #1043, opened 2023-06-22
Ranged for syntax in PVL

bobismijnnaam:
This PR adds ranged for syntax to the PVL frontend. This loop matches the use case where the bounds of the loop do not depend on the state changed within the loop, and hence strict loop bounds can be added to the contract.


src/rewrite/vct/rewrite/EncodeRangedFor.scala
```diff
@@ -0,0 +1,47 @@
...
+  override def desc: String = "Encodes ranged for as a regular for loop"
+}
+
+case class EncodeRangedFor[Pre <: Generation]() extends Rewriter[Pre] {
+  val bounds = ScopedStack[(RangedFor[Pre], Expr[Post])]()
+
+  override def dispatch(stat: Statement[Pre]): Statement[Post] = stat match {
+    case rf @ RangedFor(iv @ IterVariable(iVar, from, to), contract, body) =>
+      implicit val o = iVar.o
+      val i = Local[Post](anySucc[Variable[Post]](iVar))(iVar.o)
```
pieter-bos, line 21: it should be `succ`, not `anySucc`. `anySucc` is only for references whose type is broader than any declaration kind, e.g. `Result(_: Ref[ContractApplicable])`.

pieter-bos (changes requested):

pieter-bos (approved):

PR #1049, opened 2023-06-30
[VCLLVM] Add support for pure functions

Drevanoorschot:
This PR adds support to declare pure functions in LLVM IR named metadata nodes.
For example:
```llvm
!VC.global = !{!0}
!0 = !{!"pure i32 @add_func(i32 %y, i32 %x) = add(%y, %x);"}

define i32 @add(i32 %a1, i32 %a2)
!VC.pure !{i1 true}
{
    %res = add i32 %a1, %a2
    ret i32 %res
}

define i32 @addCaller(i32 %a1, i32 %a2)
!VC.contract !{
    !"ensures icmp(eq, \result, call @add(%a1, %a2));",
    !"ensures icmp(eq, \result, call @add_func(%a1, %a2));"
    }
{
    %res = call i32 @add(i32 %a1, i32 %a2)
    ret i32 %res
}
```

Might add some more function constructs such as implications, recursion, and what not later, but the current implementation compiles and works so can be merged already.


pieter-bos (changes requested):
Can you have a look at the formatting changes again? It's polluting the blame quite a bit. Otherwise looks good!

Drevanoorschot:
The blames should be fixed now. My bad for that, just a force of habit to spam the formatting key across files :see_no_evil:. I also discovered a bug in the `findCallable` function of the LLVM resolver. That should be fixed now too.

pieter-bos (dismissed):

PR #1053, opened 2023-07-18
Veymont gen par

petravandenbos-utwente:
A VeyMont sequential program is translated to a parallel Java program.


src/col/vct/col/ast/lang/JavaParamImpl.scala
```diff
@@ -0,0 +1,9 @@
+package vct.col.ast.lang
+
+import vct.col.ast.JavaParam
+import vct.col.print.{Ctx, Doc, Text}
+
+trait JavaParamImpl[G] { this: JavaParam[G] =>
+  override def layout(implicit ctx: Ctx): Doc = Text(t + " " + name)
```
pieter-bos, line 7: `t.show <+> name`

src/col/vct/col/ast/type/TUnionImpl.scala
```diff
@@ -1,7 +1,9 @@
 package vct.col.ast.`type`
 
 import vct.col.ast.TUnion
+import vct.col.print.{Ctx, Doc}
 
 trait TUnionImpl[G] { this: TUnion[G] =>
+  override def layout(implicit ctx: Ctx): Doc =  Doc.spread(types)
```
pieter-bos, line 7: `Doc.fold(types)(_ <+> "|" <+> _)`

pieter-bos (changes requested):
Some small changes, also if you would eventually like to pass this into vercors again (?) we should consider the places where blame is currently set to `null`.

pieter-bos (approved):

PR #1063, opened 2023-09-25
Support for SYCL's basic and nd-range kernels

Ellen-Wittingen:
This pull request adds support for SYCL's basic and nd-range kernels.

It supports the creation of new kernels with SYCL's `submit` method:
```
void main() {
  sycl::event myEvent = myQueue.submit(
    [&](sycl::handler& cgh) {
      cgh.parallel_for(sycl::range<3>(6,4,2),
      /*@
        requires it.get_id(0) > -1;
        ensures it.get_linear_id() > -1;
      */
      [=] (sycl::item<3> it) {
        int a = it.get_id(0);
        int b = it.get_linear_id();
        int c = it.get_range(0);
        //@ assert c == 6;
      });
    }
  );
  int a = 5;
  myEvent.wait();
}
```
Internally this block of code transformed into a method with a parallel block inside it, which holds the lambda function and contract written in the `parallel_for` parameters. Then `sycl::event myEvent` is assigned to a method call to the newly created method. This creates the following AST (after the language-specific rewrite pass):
```
requires (\forall* int GLOBAL_ID_0, int GLOBAL_ID_1, int GLOBAL_ID_2;
    GLOBAL_ID_0 \in {0 .. 6} && GLOBAL_ID_1 \in {0 .. 4} && GLOBAL_ID_2 \in {0 .. 2} ==> GLOBAL_ID_0 > -1
);
ref VERCORS_SYCL_KERNEL_WRAPPER() {
    par SYCL_BASIC_KERNEL(int GLOBAL_ID_0 = 0 .. 6, int GLOBAL_ID_1 = 0 .. 4, int GLOBAL_ID_2 = 0 .. 2)
    requires GLOBAL_ID_0 > -1; {
        int a;
        a = GLOBAL_ID_0;
        int b;
        b = GLOBAL_ID_2 + GLOBAL_ID_1 * 2 + GLOBAL_ID_0 * 4 * 2;
        int c;
        c = 6;
        assert c == 6;
    }
}
void main() {
    ref myQueue;
    ref myEvent;
    myEvent = VERCORS_SYCL_KERNEL_WRAPPER();
    int a;
    a = 5;
    inhale (\forall* int GLOBAL_ID_0, int GLOBAL_ID_1, int GLOBAL_ID_2;
        GLOBAL_ID_0 \in {0 .. 6} && GLOBAL_ID_1 \in {0 .. 4} && GLOBAL_ID_2 \in {0 .. 2} ==>
        GLOBAL_ID_2 + GLOBAL_ID_1 * 2 + GLOBAL_ID_0 * 4 * 2 > -1
    );
}
```
In SYCL, the host code does not wait on the submitted kernel to end before continuing. Therefore, the postcondition is not put in the generated parallel block, but is instead inhaled when `myEvent.wait()` is called, which makes the host wait on the kernel code to end.

The methods of SYCL's `item` and `nd_item` which return the current (linear) thread id and range are also transformed to the actual calculation of the work-item's (linear) id or range using the parblock variables.

Examples / tests for SYCL's kernels can be found in `examples/concepts/sycl/kernels/` and `test/main/vct/test/integration/examples/SYCLSpec.scala`.

Lambda functions are only allowed in SYCL methods, if they are used anywhere else in c++ VerCors will throw an error.

The comments on the [previous pull request](https://github.com/utwente-fmt/vercors/pull/1052) have also been addressed. The implementation of namespaces has changed, as I realized I only needed to be able to resolve methods inside them: now in the parsing stage the methods nested inside one or more namespaces get the names of the namespaces added in front of their name. This means that in the resolution stage, the methods can just be found by checking for a matching name, instead of the nested search it was before.


pieter-bos:
I hadn't started reviewing this yet so it's no problem, but just curious why it was closed?

PR #1095, opened 2023-10-27
VeyMont: `seq_run`, `:=`, structure checks and initial sequential program tranformation

bobismijnnaam:
This PR refactors the following:

- The `run` method of a `seq_program` is now a separate AST node.
- Endpoint field assignment is now a separate statement.
- The structure checks are integrated into the per-pass checking system.

A first implementation is made for encoding of trivial parallel VeyMont programs.


bobismijnnaam:
Failing tests are failing because of #1098.

src/col/vct/col/typerules/CoercingRewriter.scala
```diff
@@ -1672,6 +1672,8 @@ abstract class CoercingRewriter[Pre <: Generation]() extends AbstractRewriter[Pr
       case CommunicateX(r,s,t,a) => CommunicateX(r,s,t,a)
       case PVLCommunicate(s, r) => PVLCommunicate(s, r)
       case Communicate(r, s) => Communicate(r, s)
+      case PVLSeqAssign(r, f, v) => PVLSeqAssign(r, f, v)
+      case SeqAssign(r, f, v) => SeqAssign(r, f, v)
```
pieter-bos, line 1676: it seems suspicious that there is no typing rule here

pieter-bos (changes requested):

src/col/vct/col/typerules/CoercingRewriter.scala
bobismijnnaam: Ah, yeah

src/col/vct/col/typerules/CoercingRewriter.scala
```diff
@@ -1670,8 +1673,22 @@ abstract class CoercingRewriter[Pre <: Generation]() extends AbstractRewriter[Pr
       case w @ WandPackage(expr, stat) => WandPackage(res(expr), stat)(w.blame)
       case VeyMontAssignExpression(t,a) => VeyMontAssignExpression(t,a)
       case CommunicateX(r,s,t,a) => CommunicateX(r,s,t,a)
-      case PVLCommunicate(s, r) => PVLCommunicate(s, r)
-      case Communicate(r, s) => Communicate(r, s)
+      case PVLCommunicate(s, r) if r.fieldType == s.fieldType => PVLCommunicate(s, r)
```
bobismijnnaam, line 1676: @pieter-bos I think we could support coercions relatively easily, but I'd like to postpone that feature for now. Is this the right way of encoding that the types must be exactly the same? Also, two practical problems: 1) the error is mega ugly, it prints the entire program on error, and 2) this error is structured the same way as for the assign statement, but it results in a crash instead of an error code in the test suite. Maybe we can discuss this on tuesday :)

pieter-bos: Maybe have a look at `coerceYields` (although I really want to get rid of that at some point)

bobismijnnaam: Okay, will do. I'm assuming you're otherwise okay with how I handled the type equality.

PR #1100, opened 2023-11-03
Added support for SYCL's accessors and declarators.

Ellen-Wittingen:
Added support for SYCL's accessors and declarators, so now data on the host can be copied to kernels and written back to the host when the kernels are done. An example of the supported usage of accessors and declarators can be seen below (taken from `examples/concepts/sycl/accessors/GetKernelResult.cpp`):
```
#include <sycl/sycl.hpp>

/*@ 
  requires \pointer(a, 10, write);
*/
void main(int* a) {
	sycl::queue myQueue;

  {
    sycl::buffer<int, 1> aBuffer = sycl::buffer(a, sycl::range<1>(10));

    myQueue.submit(
      [&](sycl::handler& cgh) {
        
        sycl::accessor<int, 1> a_accessor = sycl::accessor(aBuffer, cgh, sycl::read_write);

        cgh.parallel_for(sycl::range<1>(10),
          /*@
            context it.get_id(0) < a_accessor.get_range().get(0);
            context Perm(a_accessor[it.get_id(0)], write);
            ensures a_accessor[it.get_id(0)] == 10;
          */
          [=] (sycl::item<1> it) {
            a_accessor[it.get_id(0)] = 10;
          }
        );
      }
    );
  } // Leaving scope, so automatically wait on all running kernels to finish and write data back to the buffers

  //@ assert (\forall int i; i >= 0 && i < 10; a[i] == 10);

}
```


test/main/vct/test/integration/examples/SYCLSpec.scala
```diff
@@ -4,14 +4,32 @@ import vct.test.integration.helper.VercorsSpec
...
+  vercors should error withCode "syclKernelRangeInvalid" example "concepts/sycl/kernels/NonDivisibleNDRange.cpp"
+  vercors should error withCode "syclIncorrectParallelForLambdaArgument" example "concepts/sycl/kernels/NonMatchingRangeItem.cpp"
+  vercors should error withCode "syclIncorrectParallelForLambdaArgument" example "concepts/sycl/kernels/NonMatchingRangeItemDimensions.cpp"
+  vercors should error withCode "syclNoExtraCodeInCommandGroup" example "concepts/sycl/kernels/NonSYCLCodeInCommandGroup.cpp"
   vercors should fail withCode "syclItemMethodPreFailed" using silicon example "concepts/sycl/kernels/TooHighKernelDimension.cpp"
-  vercors should error withCode "kernelForkPre" example "concepts/sycl/kernels/ZeroNDRange.cpp"
+  vercors should error withCode "syclKernelRangeInvalid" example "concepts/sycl/kernels/ZeroNDRange.cpp"
+
+  vercors should error withCode "syclAccessorInsufficientReferencePermission" example "concepts/sycl/accessors/AccessorGetRangeDimensionOutOfBounds.cpp"
+//  vercors should verify using silicon example "concepts/sycl/accessors/AllAccessModes.cpp"
```
Ellen-Wittingen, line 19: I commented out these two tests because they require the VerCors `--no-infer-heap-context-into-frame` flag to verify without VerCors throwing an error. And I do not know of a way to specify in the test which VerCors flags should be enabled.

src/col/vct/col/ast/lang/CPPLocalImpl.scala
```diff
@@ -32,5 +33,5 @@ trait CPPLocalImpl[G] { this: CPPLocal[G] =>
   }
 
   override def layout(implicit ctx: Ctx): Doc = Group(Text(name) <>
-    (if (genericArg.isDefined) (Text("<") <> Text(genericArg.get.toString) <> Text(">")) else Text("")))
+    (if (genericArgs.nonEmpty) (Text("<") <> Text(genericArgs.mkString(", ")) <> Text(">")) else Text("")))
```
pieter-bos, line 36: `Doc.args(genericArgs)`

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -728,6 +1132,183 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre]) extends L
...
+
+    // Wait for SYCL kernels to finish executing
+    var additionalStatements: Seq[Statement[Post]] = syclRunningKernelsSuccessor.pop().map(tuple => syclKernelTermination(tuple._1, tuple._2)).toSeq
+
+
+    // Destroy all buffers and copy their data back to host
+    additionalStatements =  additionalStatements ++ syclBufferSuccessor.pop().map(tuple => destroySYCLBuffer(tuple._2, scope)).toSeq
+
+    Scope[Post](result.locals, Block[Post](result.body +: additionalStatements)(result.body.o))(scope.o)
+  }
```
pieter-bos, line 1266: I'm guessing this normally happens with the RAII pattern via destructors? Do you think it is infeasible to support destructors generally? I'm a bit worried about `CPPScope` because of two things:

* The scoping part of a `CPPScope` is not meaningfully different from `Scope`, so changes to `Scope` logic should likely be reflected in `CPPScope`?
* I doubt that destructors are always related to the scope in the way I read it here, surely you can move the values or delete them earlier somehow (but I'm far from a c++ expert).

Idea: continue to use scope, and introduce a node `CppLifetimeScope(body: Statement)` that analyses what destructors to call (or emulate for sycl stuff), this at least decouples the concerns of what variables you're allowed to name and what resources need to be cleaned up.

test/main/vct/test/integration/examples/SYCLSpec.scala
pieter-bos: We should probably think about adding support for that, but can you also report the error as an issue please?

pieter-bos (changes requested):

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
Ellen-Wittingen: > I'm guessing this normally happens with the RAII pattern via destructors?

Yes, indeed. For SYCL buffers and event objects, some extra actions are performed in their destructors, other than releasing the associated memory, such as copying data back and waiting on kernels to finish running. I believe destructors can also be called in other ways than the assigned variable going out of scope (but I have also not worked with destructors in C++ before), but for the SYCL implementation it was enough to only support 'invoking' them when going out of scope.

> Idea: continue to use scope, and introduce a node CppLifetimeScope(body: Statement) that analyses what destructors to call (or emulate for sycl stuff), this at least decouples the concerns of what variables you're allowed to name and what resources need to be cleaned up.

I implemented this now, it is indeed a better solution.

pieter-bos:
I've not merged this yet because you are failing this test: <https://github.com/utwente-fmt/vercors/actions/runs/6768903759/job/18394515499?pr=1100>

Ellen-Wittingen:
> I've not merged this yet because you are failing this test: https://github.com/utwente-fmt/vercors/actions/runs/6768903759/job/18394515499?pr=1100

You mean this?

> ExampleCoverage:
Not tested: examples/concepts/sycl/accessors/AllAccessModes.cpp
Not tested: examples/concepts/sycl/accessors/GetKernelResult.cpp
\- should cover all examples in the examples directory *** FAILED ***
  The test suite does not have a test entry that processes the above files. (ExampleCoverage.scala:70)

These two tests are commented out in SYCLTest, because they need a flag to pass, see [my comment here](https://github.com/utwente-fmt/vercors/pull/1100#discussion_r1381744780)

I see two options to solve this:
- Add the tests, but write them to expect an error for now `vercors should error withCode 'CODE' example "concepts/sycl/accessors/AllAccessModes.cpp"` and add a comment about the issue above these tests.
- Removing these tests from the `examples` directory.

These tests are quite important, as they test some core functionality of SYCL's data accessors. I also expect to add more example files for the next research questions which will probably have the same issue. So I think I would prefer the first option, what do you think?

pieter-bos:
Right, sorry. Let's go for the first solution right now, I'll think about adding flags to test fixtures.

Ellen-Wittingen:
I added the tests. I did have to add the line `case Left(err: BlameUnreachable) if code.equals("unreachable:schematic") && err.message.equals("schematic") =>` to `VercorsSpec` to allow the tests to "pass"

PR #1118, opened 2023-12-07
Redo helpers: make a proper meta-model of Node.scala

pieter-bos:
* Analyse Node.scala before the build is constructed
* Define code generation as fine-grained tasks in mill
* Ensure we only generate node helpers whose definition has changed
* Define most operations directly on `Impl` traits (over extension methods via implicit classes)


mill-build/analysis/src/StatAnalysis.scala
```diff
@@ -0,0 +1,169 @@
...
+  def is[M <: Mod](mods: Seq[Mod])(implicit tag: ClassTag[M]): Boolean =
+    mods.collectFirst { case _: M => () }.nonEmpty
+
+  def getDeclaration(rawStat: RawStatAnalysis.RawStat): Result[Decl] = Try(rawStat.blame) {
+    val types = TypeAnalysis(typeLookup, scope = rawStat.name.initName)
+
+    val isConcrete = !rawStat.isTrait && !is[Mod.Abstract](rawStat.mods)
+
+    val modConditions =
+      if (isConcrete) assertPure(true /*is[Mod.Final](rawStat.mods)*/ , rawStat.blame, "Node definitions must be final")
```
bobismijnnaam, line 27: This assertPure doesn't do anything, right?

bobismijnnaam: (Same a few lines below, there's another one)

bobismijnnaam:
- Line 531 in build.sc gives the following warning:

```
[#5] [warn] /home/bobe/Workbench/Projects/vercors_dev/build.sc:531:43: reference to scalaVersion is ambiguous;
[#5] [warn] it is both defined in the enclosing object silicon and inherited in the enclosing object buildInfo as method scalaVersion (defined in trait ScalaModule)
[#5] [warn] In Scala 2, symbols inherited from a superclass shadow symbols defined in an outer scope.
[#5] [warn] Such references are ambiguous in Scala 3. To continue using the inherited symbol, write `this.scalaVersion`.
[#5] [warn] Or use `-Wconf:msg=legacy-binding:s` to silence this warning. [quickfixable]
[#5] [warn]           BuildInfo.Value("scalaVersion", scalaVersion()),
[#5] [warn]                                           ^
[#5] [warn] one warning found
```

Maybe you can fix that if you have the time

- folder colhelpers has to be deleted?
- few "assertPure" expressions have "true", should these be filled in or am I misunderstanding their side effects?
- rootNode in HierarchyAnalysis.scala is not used
- Analysis.scala: maybe use the same option library we use in vercors? Robustness is nice but if it takes more than a few lines of code, it might not be worth the effort

PR #1121, opened 2023-12-12
Added support for SYCL's local accessors plus minor changes

Ellen-Wittingen:
Local accessors, which provide each workgroup with their own unique array of data, can now be declared and used. An example of the supported usage of local accessors can be seen below (taken from `examples/concepts/sycl/addressSpaces/LocalAccessorUsage.cpp`):
```c++
#include <sycl/sycl.hpp>

void test() {
  sycl::queue myQueue;

  myQueue.submit(
    [&](sycl::handler& cgh) {
      // Create local accessor to an array of 5 uninitialized integers for each work-group
      sycl::local_accessor<int, 1> a_local_acc = sycl::local_accessor<int, 1>(sycl::range<1>(5), cgh);

      cgh.parallel_for(sycl::nd_range<1>(sycl::range<1>(20), sycl::range<1>(5)),
        /*@
          context it.get_local_id(0) < a_local_acc.get_range().get(0);
          context Perm(a_local_acc[it.get_local_id(0)], write);
          ensures a_local_acc[it.get_local_id(0)] == 10;
        */
        [=] (sycl::nd_item<1> it) {
          // Use the local accessor (syntax and required pre- / post-conditions same as for data accessors)
          a_local_acc[it.get_local_id(0)] = 10;
        }
      );
    }
  );
}
```

Minor changes that were also made:
- Users can now use expressions (instead of only constant integers) in the range parameter of `parallel_for(range, kernel lambda)` inside command groups.
- The methods of `item` or `nd_item` are no longer directly inserted into COL. Calls to them are now transformed to field references or calls to the `linearize` methods in the header file. 
- Because the range parameter of `parallel_for(range, kernel lambda)` can now be an expression, range fields were added to the generated kernel class. This made the verifier think that the item methods, which now use those fields instead of constant integers, were no longer injective. So conditions were added that prove the injectivity of item methods so they can be used in statements such as `context Perm(a_local_acc[it.get_local_id(0)], write);`.
- Moved buffer and accessor constructors from the header file to Scala code. This made handling generic arguments easier and allows all possible types to be used in their generic arguments instead of only primitive types.
- Added support for for-loops with all initializers empty: `for(; ;) {}` (I had simply forgotten to do this, implemenation was copied from C implementation)


Ellen-Wittingen:
I expect to make one more final merge request in the first half of January with some bigger examples and maybe some bugfixes if I find any more bugs. 

So this is the last merge request to add new features.

pieter-bos (approved):

PR #1124, opened 2023-12-21
Quantifier triggers are now also processed when they are location

sakehl:
I discovered that when you write:

```java
    requires x >= 0 && x < 4 && y >= 0 && y < 3;
    ensures 0 <= \result && \result < 12;
pure int ind(int x, int y) = (y * 4) + x;
  
  context inp != null && inp.length == 3*4;
  context (\forall* int xf, int yf; ((0 <= xf) && (xf < 4)) && ((0 <= yf) && (yf < 3)); Perm({:inp[ind(xf, yf)]:}, 1\2));
int main(int[] inp){
    return 0;
}

```

The trigger `{:inp[ind(xf, yf)]:}` was not processed correctly, and warnings were given.

This should fix that.


pieter-bos:
Hey, good plan! Could you have a look at all the usages of the other Locations? Generally rewrites expect that a kind of `Location` only occurs at the top, and this makes it nested. For example:

* `InlineApplicables` filters for `PredicateLocation` here: https://github.com/utwente-fmt/vercors/blob/78f76cbfc0e7543ee50a4f79d5dbe636a457293d/src/rewrite/vct/rewrite/InlineApplicables.scala#L258-L262
* If it occurs differently than that, we crash: https://github.com/utwente-fmt/vercors/blob/78f76cbfc0e7543ee50a4f79d5dbe636a457293d/src/rewrite/vct/rewrite/InlineApplicables.scala#L171-L175

src/rewrite/vct/rewrite/InlineApplicables.scala
```diff
@@ -261,6 +261,14 @@ case class InlineApplicables[Pre <: Generation]() extends Rewriter[Pre] with Laz
     case Perm(loc @ InstancePredicateLocation(pred, obj, args), WritePerm()) if pred.decl.inline =>
       dispatch(InstancePredicateApply(obj, pred, args, WritePerm()(loc.o))(loc.o))
 
+    case Perm(InLinePatternLocation(loc @ PredicateLocation(pred, args), pat), WritePerm()) if pred.decl.inline =>
+      dispatch(PredicateApply(pred, args, WritePerm()(loc.o))(loc.o))
+//      TODO: Or should I do:
+//      dispatch(InlinePattern(PredicateApply(pred, args, WritePerm()(loc.o))(loc.o))(loc.o))
```
sakehl, line 267: So this one I wasn't sure about, what do you think?

sakehl:
So I did a find all on everything that seems to have 'Location' as part of its name. And I think only in the case of InlineApplicable a fix was needed.

In some other cases I could add it (for example line 71 of vct/rewrite/Explode.scala), but the AST in those passes should not need it, since `ExtractInlineQuantifierPatterns` already removed any occurrences of `InLinePatternLocation`.

src/rewrite/vct/rewrite/InlineApplicables.scala
pieter-bos: I think the commented alternative is the least surprising option.

sakehl:
Should be fine as is now.

PR #1127, opened 2024-01-10
Quick fixes

sakehl:
Had some old changes still laying around, which help fix things / make the output a bit nicer.


src/col/vct/col/print/Ctx.scala
```diff
@@ -30,8 +31,10 @@ case class Ctx(
       namer.finish.asInstanceOf[Map[Declaration[_], String]]
     })
 
-  def name(decl: Declaration[_]): String =
-    names.getOrElse(decl, s"${decl.o.getPreferredNameOrElse().ucamel}_${decl.hashCode()}")
+  def name(decl: Declaration[_]): String = {
+    val hash = if(withoutHash) "" else "_${decl.hashCode()}"
+    names.getOrElse(decl, s"${decl.o.getPreferredNameOrElse().ucamel}${hash}")
```
pieter-bos, line 36: Rendering the topmost nodes (like `Program`) never has any unassigned names, because all the declarations are contained in them. User-facing output should be rendered in the context of the program. Thus: really the only situation in which declarations are not named is if we debug them for output, and I think it is important that inequal declarations with the same name are distinguishable across different `toString` invocations. All to say: I think for the trigger warning here this is not a good solution. Probably we should render a fragment from the origin instead?

pieter-bos: If the origin context is too often inaccurate (say for parallel block generated quantifiers) then we can also consider doing something like `program.highlight(e).messageInContext("contains no triggers")` or so, ensuring everything is named correctly.

sakehl: It is mostly that either with the origins I tried to display it, and then either:
1. with `messageInContext` an enormous amount of context was shown for each warning, which is unwanted.
2. Other context options a tried, cut-out the middle part of the forall. So you saw something like:
`[Warning] blablabla: (\forall int i, int j, int k, 0 <= i ....... , write));' or something like that, which really does not give you great information on how this forall looked like. Especially for those generated from parallel blocks.

I'd have to lookup what I did again here, I'll get back to that.

Also, I'll try `program.highlight(e).messageInContext("contains no triggers")` to see if it satisfactory.

sakehl: Also, thanks for the review!

sakehl: I've changed this in such a way that we have the `inlineContext` method now has an option to compress it's output or not.

Now this example program:
```c
int main(){
    int x[10];
    int n = 10;
    /*@
    loop_invariant 0 <= i && i <= 10;
    loop_invariant (\forall* int j; 0 <= j && j < 10; Perm(&x[j], write));
    loop_invariant (\forall int j, int k; 0 <= j && j < n && 0 <= 0 && k < 2  ; x[10*j+k] == j);
    // loop_invariant (\forall int j; 0 <= j && j < i; x[j] == j);
    @*/
    for(int i = 0; i < 10; i++){
        int x = i;
    }
}
```
Gives the following warning:
`[WARN] The binder 'trigger_warning.c:7-:7`:`(\forall int j, int k; 0 <= j && j < n && 0 <= 0 && k < 2  ; x[10*j+k] == j) contains no triggers' `

which works for me

Issue #1141, opened 2024-01-23
Cannot use "NULL" as struct pointer


ArmborstL:
The C example below has a struct containing a struct pointer. Initialising that pointer to `NULL` does not work.
```c
#include <stdlib.h>

struct nested {
  struct nested *inner;  
};

void main() {
  int *ip = NULL;                               // works
  double *dp = NULL;                            // works
  struct nested *np = NULL;                     // works
  // np = (struct nested*) NULL;                // causes "This cast is not supported"
  np = (struct nested*) malloc(sizeof(struct nested));
  // np->inner = NULL;                          // causes "Expected the type of this expression to be `pointer<struct Unknown_35326835>`, but got ??TNull??()"
  // np->inner = (struct nested*) NULL;         // causes "This cast is not supported"
}
```
The first attempt of just assigning `NULL` to `np->inner` on line 13 fails with the error _"Expected the type of this expression to be `pointer<struct Unknown_1912014070>`, but got ??TNull??()."_ pointing to the `NULL` element.
Consequently, I tried to cast explicitly (line 14), but this causes the error _"This cast is not supported"_ pointing to the cast expression `(struct nested*) NULL`.

Interestingly enough, the assignment on line 10 works (i.e. when the assignment target `*np` is not within a struct), but line 11 (same, but with explicit cast) also says _"This cast is not supported"_.

Maybe @sakehl knows where this is coming from?

ArmborstL:
Note: It's clear that the _"This cast is not supported"_ comes from https://github.com/utwente-fmt/vercors/blob/3ad6fea2a3e33f603d4a35773e307bea615ab1df/src/rewrite/vct/rewrite/lang/LangCToCol.scala#L297, where C casts are handled, and only a very restricted number of cases is supported. I'm more confused by the implicit cast working in some cases, but not always...

Edit: Thanks Lars for the quick reply :) I think casting `NULL` to any pointer type might be a reasonable addition to the limited list of allowed casts, right? Thanks for looking into the implicit one :)

sakehl:
We could allow NULL to any pointer type I suppose.

Anyway for the other thing, if you add:

`case (TNull(), CTPointer(target)) => CoerceNullPointer(target)`

to line 71 of vct/col/typerules/CoercionUtils.scala
it should work.

I'll try to look at CCast if it reasonable to add it.

sakehl:
I've made a quick pull request to fix this, and added your example as a test. Should work now.

PR #1142, opened 2024-01-23
Fix for #1141

sakehl:
Fixes for https://github.com/utwente-fmt/vercors/issues/1141

### DIFF
--- a/test/main/vct/test/integration/examples/CSpec.scala
+++ b/test/main/vct/test/integration/examples/CSpec.scala
@@ -379,13 +379,13 @@ class CSpec extends VercorsSpec {
     };
 
     void main() {
-      int *ip = NULL;                               // works
-      double *dp = NULL;                            // works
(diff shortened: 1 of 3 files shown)


test/main/vct/test/integration/examples/CSpec.scala
```diff
@@ -369,4 +369,23 @@ class CSpec extends VercorsSpec {
...
+
+    struct nested {
+      struct nested *inner;  
+    };
+
+    void main() {
+      int *ip = NULL;                              
+      double *dp = NULL;                           
+      struct nested *np = NULL;                    
+      // np = (struct nested*) NULL;
```
ArmborstL, line 385: This line should not be commented out any longer, right? (same for the two below)

sakehl: Thanks you are right, fixed.

PR #1149, opened 2024-02-06
Fix bug which caused VCLLVM to get stuck

superaxander:
When processing a large LLVM-IR file we get stuck because VCLLVM is waiting for someone to read its output stream because the buffer has filled up. Instead of first waiting for the program to finish we can read the output into a ByteArrayOutputStream while its running.

### DIFF
--- a/src/parsers/vct/parsers/ColLLVMParser.scala
+++ b/src/parsers/vct/parsers/ColLLVMParser.scala
@@ -11,7 +11,9 @@ import vct.col.ref.Ref
 import vct.parsers.transform.{BlameProvider, LLVMContractToCol, OriginProvider}
 import vct.result.VerificationError.{SystemError, Unreachable, UserError}
 
-import java.io.{InputStreamReader, StringWriter}
+import java.io.IOException
+import java.nio.charset.StandardCharsets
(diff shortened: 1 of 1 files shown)


pieter-bos:
Ah, nice catch! Shall we use a similar approach to [this](https://github.com/utwente-fmt/vercors/blob/bfd486d842d92d08bdde1d2156b9faca66115872/src/parsers/vct/parsers/ColCParser.scala#L59-L69)? That is: do the transfer action in a different thread started prior to the process wait. That way the data is not duplicated in memory first.

superaxander:
I don't think it's easy to avoid duplicating in memory here. In this case we need to end up with a buffer containing all the bytes which will be the array underlying the ByteArrayOutputStream. An `InputStream#transferTo` call would use a similar buffer and actually this code can be simplified a bit more by calling `input.transferTo(baos)` in the loop. The CodedInputStream does use another buffer because it will be reading from the baos but that's relatively small.

Of course we can run just call `Program.parseFrom(process.getOutputStream)` and catch any exceptions that throws to turn that into an LLVMParseError and read the error stream instead of waiting for the process to finish ourselves. In that case we don't even need another thread.

PR #1154, opened 2024-02-23
Generics for PVL classes

bobismijnnaam:
Before merging this PR, please check the following:

- [x] I made sure the wiki is updated in accordance with the changes in this PR. For example, syntax changes, semantics changes, VerCors flags changes, etc.
 
(I will write a wiki page for generics later)

# PR description

Basic generics should now work. It is mostly an extension of the generics system that was already present for applicables, with a little optimization such that only the generic instantiation of each class is fully verified.


src/col/vct/col/ast/lang/pvl/PVLInvocationImpl.scala
```diff
@@ -1,17 +1,30 @@
...
+          tcls.instantiate(returnType)
+        }
+      }.getOrElse(returnType)
+    case RefInstanceMethod(decl) =>
+      val returnType = decl.returnType.particularize(decl.typeArgs.zip(typeArgs).toMap)
+      obj.flatMap { e =>
+        e.t.asClass.map { tcls =>
+          tcls.instantiate(returnType)
+        }
+      }.getOrElse(returnType)
```
pieter-bos, line 27: You don't need to be conservative here: `t` may not be called before the node is checked, so please use `e.t.asClass.get`.

pieter-bos: Also if the `flatMap` is for the case that the object is implicitly `this`, maybe put that in a comment :)

src/col/vct/col/ast/type/typeclass/TypeImpl.scala
```diff
@@ -44,7 +44,7 @@ trait TypeImpl[G] extends TypeFamilyOps[G] { this: Type[G] =>
       override def succProvider: SuccessorsProvider[G, G] = IdentitySuccessorsProvider
 
       override def dispatch(t: Type[G]): Type[G] = t match {
-        case TVar(Ref(v)) => substitutions(v)
+        case t @ TVar(Ref(v)) => substitutions.get(v).getOrElse(t)
         case other => rewriteDefault(other)
       }
```
pieter-bos, line 49: I understand why you do this, but I think it's a bit bug-prone. Do you think it's possible to add explicit identity mappings where you depend on this behaviour?

src/col/vct/col/typerules/Types.scala
```diff
@@ -51,7 +51,8 @@ object Types {
     case (TType(left), TType(right)) =>
       TType(leastCommonSuperType(left, right))
 
-    case (TClass(left), TClass(right)) =>
+    case (TClass(left, Seq()), TClass(right, Seq())) =>
+      // TODO (RR): Not sure how generics factor into this part... See also: next match statement
```
pieter-bos, line 55: `transSupportArrows` should be defined on `TClass` and return a `Seq[(TClass[G], TClass[G])]`. We don't support {contra|co}variance yet (other than explicitly for axiomatic collections), so otherwise the type has to be equal.

src/col/vct/col/typerules/Types.scala
```diff
@@ -63,11 +64,11 @@ object Types {
...
         case Nil => TAnyClass()
-        case Seq(t) => TClass(t.ref)
-        case other => TUnion(other.map(cls => TClass(cls.ref)))
+        case Seq(t) => TClass(t.ref, Seq())
+        case other => TUnion(other.map(cls => TClass(cls.ref, Seq())))
       }
 
-    case (TClass(_), TAnyClass()) | (TAnyClass(), TClass(_)) =>
+    case (TClass(_, _), TAnyClass()) | (TAnyClass(), TClass(_, _)) =>
       TAnyClass()
```
pieter-bos, line 72: Seems correct

src/main/vct/main/stages/Transformation.scala
```diff
@@ -122,17 +123,23 @@ class Transformation
...
         }
 
         onAfterPassKey.foreach {
           case (key, action) => if (pass.key == key) action(result)
         }
 
+        result.tasks.map(_.program).flatMap(program => program.check.map(program -> _)) match {
+          case Nil => // ok
+          case errors => throw TransformationCheckError(pass, errors)
+        }
```
pieter-bos, line 141: Generally I would expect that attempting to print a malformed ast is less diagnostically useful than analyzing the reason the ast is malformed, did you have a case where that is not true?

src/rewrite/vct/rewrite/exc/EncodeBreakReturn.scala
```diff
@@ -174,10 +174,10 @@ case class EncodeBreakReturn[Pre <: Generation]() extends Rewriter[Pre] {
 
               if(needReturn(method)) {
                 val returnField = new InstanceField[Post](dispatch(method.returnType), Nil)(ReturnField)
-                val returnClass = new Class[Post](Seq(returnField), Nil, tt)(ReturnClass)
+                val returnClass = new Class[Post](Nil, Seq(returnField), Nil, tt)(ReturnClass)
```
pieter-bos, line 177: Aww, surely you saw the opportunity here :D

pieter-bos (changes requested):

src/rewrite/vct/rewrite/exc/EncodeBreakReturn.scala
bobismijnnaam: Oh! That's a good idea, also a good way to indirectly test the system much more thoroughly. I'll save that for later.

src/main/vct/main/stages/Transformation.scala
bobismijnnaam: Oh, I accidentally left that there, I should put it back. Not entirely sure but I think there were one or two cases where I just got a "no successor here" or "the following key does not exist" (from succmap) error, and then it was nice to look at the output and find the ??brokenref??s. Once I saw them in context it was easy to identify the code that produced them. I think there were also a few typechecking problems caused by veymont that were nice to diagnose like this.

src/col/vct/col/ast/type/typeclass/TypeImpl.scala
bobismijnnaam: Turns out I was just not including all type variable assignments up front - once you do, identity mappings should not be necessary if I understand PVL's ad-hoc generics type system correctly.

PR #1156, opened 2024-02-26
Vector support in C, OpenCL and CUDA + Vector ADT in PVL

sakehl:
* I've added support for vectors in C, OpenCL, CUDA and added a custom ADT in PVL.
* PVL/use in annotations:
  * a vector type: `vector<int,4> a4`
  * A vector constructor: `a4 = vector<int>{1,2,3,4};`
  * Vector can be accessed with `[]`: `a4[0]`
  * Support for operators `+`, `-`, `/`, `*`, `%"`, if the vector has int/float type: `a4 + a4`
  * Support for `==`, `!=` operators:
    * I've made the choice for this to return vectors of Type Int. This is because in C the "==" the operators behaves like this on vectors. And now it is the same behavior. I'm unsure if this is the best approach though. Anyway for this happens:
    `vector<int>{1,2,3,4} == vector<int>{1,0,-2,4}` evaluates to `vector<int>{1,0,0,1}`.
   Maybe we should return `vector<bool>` here instead?
* C:
  * Types can now be defined as a vector: `typedef int v4si __attribute__ ((vector_size (sizeof(int)*4)));` 
    * For this I added `typedef` in general.
  * A Vector can be initialized: `v4si res = {1,2,3,4}`
  * Can be indexed `res[0] == 1`
  * Can be updated `res[0] = 2`
    * Since I encode vectors as ADT, this is a shorthand to update the whole vector.
  * Also has support for operators as above
* OpenCL: OpenCL uses vectors slighly different than C. Mostly in constructing an accessing them.
  See [here](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#built-in-vector-data-types), [here](  https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#vector-data-load-and-store-functions) and [here](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#vector-components)
  * So I use a custom _VerCors internal_ type for them: `typedef __opencl_vector_type__(int, 2) int2;`
  * Initialize them as: `int2 x = (int2)(0,1)`.
    Note that this uses the function invocation syntax, so I do the same internally.
    This initializer can take any form of vector/scalar arguments, as long as the total scalars is as expected, e.g.:  `int4 y = (int4)(x, 2, 3)`.
  * Access/update them with fields. E.g you can write `x.xy` or `x.yy`, `y.0123` `y.even`, `y.odd` `y.lo` `y.hi`, etc.
  * The `vloadn` and `vstoren` functions. In the OpenCL standard they are defined for multiple types (float, int, etc). But this seemed to hard for now, so I only defined them for `int`.
* CUDA: Cuda actually doesn't really have _vectors_, as in, it does not define any operations working on them. see [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#built-in-vector-types).
  *  There are the types `int2` etc, but they are equivalent to structs as for as I can see: e.g. `struct int2{int x, y;};` So I've added these definitions in `vector_types.h`. And `vector_functions.h`. 
  *  CUDA seems to have vector types defined defined since doing `int* x = ...; int2* x2 = (int2*)x; x2[0];` seems to lead to one memory load in the produced assembly code , instead of two loads, if two integers are loaded like `x[0]` and `x[1]`.

## Additional comments
The first goal for me was to support the AVX/SSE libraries, like:
[`__m128i _mm_add_epi32 (__m128i a, __m128i b)`](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_add_epi32&ig_expand=89).

The problem here is that `__m128i` is defined as a `int` 64bit vector of length two. How it is actually _used_ is defined by the operators. E.g. `_mm_add_epi32` treats the type as a `int` 32bit vector of length four, so summing vectors of length 4. Whilst `_mm_add_epi64`, with the same type, treats it as summing vectors of length 2.

The type does matter, since `__m128i x = {0, 1}` is valid to load a vector of length 2. But `__m128i x = {0, 1, 2, 3}` does not give back a vector of length 4.

An idea to handle this, would be to track two types for such variables.
* Type 1: The type that the C file actually needs (since this is the type the compiler needs to compile the C file)
* Type 2: How we interpret the type in VerCors.

E.g.
```c
#include <immintrin.h>

typedef int __v4si __attribute__((__vector_size__(sizeof(int)*4)));

void test(){
  int a[4] = {0,1,2, 3};
  __m128i a4 /*@ as __v4si @*/;
  a4 =  _mm_load_si128((__m128i*) a)  /*@ as __v4si @*/;
  a4 =  _mm_add_epi32(a4, a4);
  _mm_store_si128((__m128i*) a /*@ as __v4si @*/, a4);
}
```

or something similar. 

## Wiki:
If you agree with the choices here, I'll add a wiki entry about the vector ADT.


src/col/vct/col/ast/Node.scala
```diff
@@ -622,6 +646,10 @@ final case class ResourceOfResourceValue[G](res: Expr[G])(implicit val o: Origin
 final case class ResourceValue[G](res: Expr[G])(implicit val o: Origin) extends Expr[G] with ResourceValueImpl[G]
 
 sealed trait Comparison[G] extends BinExpr[G] with ComparisonImpl[G]
+sealed trait AmbiguousComparison[G] extends Comparison[G] with AmbiguousComparisonImpl[G]
+final case class AmbiguousEq[G](left: Expr[G], right: Expr[G], vectorInnerType: Type[G])(implicit val o: Origin) extends AmbiguousComparison[G] with AmbiguousEqImpl[G]
+final case class AmbiguousNeq[G](left: Expr[G], right: Expr[G], vectorInnerType: Type[G])(implicit val o: Origin) extends AmbiguousComparison[G] with AmbiguousNeqImpl[G]
```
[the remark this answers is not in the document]
pieter-bos: Ah I missed `VectorEq`/`VectorNeq`, ok.

src/col/vct/col/typerules/CoercingRewriter.scala
```diff
@@ -451,7 +496,8 @@ abstract class CoercingRewriter[Pre <: Generation]() extends BaseCoercingRewrite
                   alt8: => T = throw IncoercibleDummy,
                   alt9: => T = throw IncoercibleDummy,
                   alt10: => T = throw IncoercibleDummy,
-                  alt11: => T = throw IncoercibleDummy) : T = {
+                  alt11: => T = throw IncoercibleDummy,
+                  alt12: => T = throw IncoercibleDummy) : T = {
```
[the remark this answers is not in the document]
pieter-bos: jeez, `AmbiguousPlus` really has that many alternatives...

src/col/vct/col/typerules/CoercingRewriter.scala
```diff
@@ -1398,10 +1475,29 @@ abstract class CoercingRewriter[Pre <: Generation]() extends BaseCoercingRewrite
...
+      case div @ VectorFloatDiv(_, _) => vectorFloatOp2(div, (l,r) => VectorFloatDiv(l,r)(div.blame))
+      case div @ VectorFloorDiv(_, _) => vectorIntOp2(div, (l,r) => VectorFloorDiv(l,r)(div.blame))
+      case minus @ VectorMinus(_, _) =>
+        vectorOp2(minus, (l,r) => VectorMinus(l,r))
+      case mod@ VectorMod(_, _) => vectorIntOp2(mod, (l,r) => VectorMod(l,r)(mod.blame))
+      case mult @ VectorMult(_, _) =>
+        vectorOp2(mult, (l,r) => VectorMult(l,r))
+      case VectorNeq(left, right) =>
+        val sharedType = Types.leastCommonSuperType(left.t, right.t)
+        VectorNeq(coerce(left, sharedType), coerce(right, sharedType))
```
pieter-bos, line 1490: `Eq`/`Neq` should be vectors no?

src/rewrite/vct/rewrite/Disambiguate.scala
```diff
@@ -105,13 +124,22 @@ case class Disambiguate[Pre <: Generation]() extends Rewriter[Pre] {
...
         else if(op.isSeqOp) SeqMember(dispatch(x), dispatch(xs))
         else throw Unreachable("AmbiguousMember must query a map, set, bag, or seq because of the type check.")
+      case cmp: AmbiguousComparison[Pre] =>
+        if(cmp.isVectorOp) cmp match {
+          case AmbiguousEq(left, right, _) => VectorEq(dispatch(left), dispatch(right))
+          case AmbiguousNeq(left, right, _) => VectorNeq(dispatch(left), dispatch(right))
+        } else cmp match {
+          case AmbiguousEq(left, right, _) => Eq(dispatch(left), dispatch(right))
+          case AmbiguousNeq(left, right, _) => Neq(dispatch(left), dispatch(right))
+        }
```
pieter-bos, line 142: Is this vector inner type in `AmbiguousEq` used for something still? It seems like a strange abstraction.

res/universal/res/adt/vector.pvl
```diff
@@ -0,0 +1,8 @@
+adt `vector`<T> {
+  pure T vector_loc(`vector`<T> v, int i);
+}
+
+    decreases;
+    requires 0 <= i;
+    requires i < 4;
+pure T v4loc<T>(`vector`<T> a, int i) = `vector`<T>.vector_loc(a, i);
```
pieter-bos, line 8: If I understand correctly `v4loc` is an unused example, right?

pieter-bos (changes requested):

src/col/vct/col/ast/Node.scala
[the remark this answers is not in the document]
sakehl: Sure I've added MapEq in disambiguate

res/universal/res/adt/vector.pvl
sakehl: True, the actual methods are generated. But I wanted an example you could just see.

src/col/vct/col/ast/expr/op/cmp/AmbiguousComparisonImpl.scala
```diff
@@ -0,0 +1,11 @@
+package vct.col.ast.expr.op.cmp
+
+import vct.col.ast.{AmbiguousComparison, TBool, TVector, Type}
+
+trait AmbiguousComparisonImpl[G] { this: AmbiguousComparison[G] =>
+  // So this is hacky, but in C languages the type of a vector comparison is vector<Int>, and here we need to make difference
+  // Between CInt and Int because of the internal VerCors type checker.
+  def vectorInnerType: Type[G]
+
+  override def t: Type[G] = if(isVectorOp) TVector[G](getVectorType.size, vectorInnerType) else TBool()
```
sakehl, line 10: So I need the vectorInnerType here, to decide to return a `vector<size, Int>` or `vector<size, CInt>`

src/rewrite/vct/rewrite/Disambiguate.scala
sakehl: See the comment I made below

Issue #1152, opened 2024-02-22
The --skip-backend flag does nothing


superaxander:
The --skip-backend command line option does nothing. The boolean it sets is never used.

PR #1157, opened 2024-02-26
Actually skip backend

sakehl:
Makes sure that the option `--skip-backend` is actually used. (Thus fixes https://github.com/utwente-fmt/vercors/issues/1152).

Additionally adds the option `dev-total-timeout`. This adds a timeout towards only the verification part of the back-end, which kills the process via Silver after a timeout. I found this useful, since with this timeout, back-end processes always get killed correctly.  In contracts, sometimes if I killed VerCors with the bash program `timeout` and some Z3 programs kept running in the background.

### DIFF
--- a/src/viper/viper/api/backend/SilverBackend.scala
+++ b/src/viper/viper/api/backend/SilverBackend.scala
@@ -56,7 +56,7 @@ trait SilverBackend extends Backend with LazyLogging {
   private def path(node: silver.Node): Seq[AccountedDirection] =
     info(node.asInstanceOf[silver.Infoed]).predicatePath.get
 
-  override def submit(colProgram: col.Program[_], output: Option[Path]): Boolean = {
+  override def submit(colProgram: col.Program[_], output: Option[Path], skipVerification: Boolean): Boolean = {
     val (silverProgram, nodeFromUniqueId) = ColToSilver.transform(colProgram)
 
     val silverProgramString =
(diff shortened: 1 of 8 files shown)


src/viper/viper/api/backend/SilverBackend.scala
```diff
@@ -96,6 +96,8 @@ trait SilverBackend extends Backend with LazyLogging {
             }
         }
     }
+    // Early out of verification
+    if(skipVerification) return true
```
pieter-bos, line 100: I think this will potentially cache the wrong result, no? I think the stage should just make sure not to call Viper instead.

sakehl: Hmm Fair I'm not sure how the caching works, so I did not take that in account.

The reasoning of putting it here, is such that you still get output with the option `--backend-file-base`, but do not start verifying. This is (Which I found useful to debug/inspect the viper file).

pieter-bos: That's fair, let's split that into two stages then :) (so Col -> Silver, Silver -[verification]-> errors). I understand if you don't want to bother with that right now, but I don't want versions around that put wrong records in the cache, so maybe you can accept no viper file in the meantime.

sakehl: Yeah that seems fair, I'll have a look if this is possible in reasonable time.

I'll move some code from the 'submit' method.

sakehl: So I split the 'submit' method into a 'transform'  and 'submit' method. Adding an type for the intermediate program after the transform method was a bit of trouble, but this should work.

PR #1159, opened 2024-02-28
Update VCLLVM (now Pallas) to LLVM 17, update to newest VerCors version, and convert more instructions to COL

superaxander:
# Summary
This PR updates VCLLVM to LLVM 17 and updates it to work with the current VerCors. Additionally more instructions converted into COL, there is support for loops, pointers, structs, and more. This also includes changes to the C and PVL frontends for pointers and ByValueClass encoding. (i.e. structs)

# Detailed list of changes
## General
- Ignore a quantifier in SimplifyNestedQuantifiers if it contains a non-inline trigger (it was erroneously rewriting a generated trigger)
- Make the ParBlockEncoder treat a unary minus as a constant operator (this was needed for one of the HaliVer examples to verify)
- Update to silicon commit https://github.com/viperproject/silicon/commit/2030e3eb63f4b1c92ddc8885f7c937673effc9bd to improve verification performance for quantified heap chunks (PR: https://github.com/viperproject/silicon/pull/860)
- Add clang-format to the commit hooks to format the C++ code before every commit
## Pallas
- Rename VCLLVM to Pallas
- Update Pallas to LLVM 17 from LLVM 15
- Switch from Pallas as a standalone application to a plugin that can be loaded by the LLVM framework
- Add support for more LLVM-IR instructions
  - AllocA (allocating variables on the stack eliminating the need for the mem2reg pass before verification)
  - GetElementPtr (indexing structs arrays and vectors, currently only with statically known offsets)
  - Load/Store (loading and storing values behind pointers)
  - SExt, ZExt, Trunc (since we only deal with mathematical integers these are translated as no-ops for now)
- Add support for more LLVM-IR types
  - Structs
  - Pointers (including some very basic function pointer support)
  - Arrays/Vectors (these are currently treated as the same)
- Add support for more LLVM-IR elements:
  - Global variables
  - Loops (Branch instructions are transformed into ifs + goto and then the goto's are replaced with actual (while) loops. This uses LLVM's loop analyses to find the header and body of the loop)
  - More constant types
  - Implicit pointer casts (based on C support described below, but with automatic type inference since LLVM omits pointer types)
- Pass in debug locations from the LLVM-IR to VerCors so that error locations in the original source files can be pointed out
## C (changes superceding #1172 and #1227)
- Change permission syntax such that it always takes a pointer except when unfolding a struct permission
  - Permission to write a field `f` of struct value `a` becomes `Perm(&a.f, write)` instead of `Perm(a.f, write)`
  - Permission to write all fields of a struct value `a` becomes `Perm(a, write)` (or if `a` is a pointer to a struct you'd have `Perm(a, write) ** Perm(*a, write)`)
- Replace struct encoding with a new one based on ADTs (called ByValueClass in the code) containing pointers (related: #1194, #1227)
  - Allow casting pointers to structs to pointers to their first element (possibly multiple layers deep), also allow casting these pointers back up
  - Allow passing around void pointers to have pointers that might not always point to the same type
  - Allow getting a pointer to a struct field
  - Ensure struct values are always treated as values instead of references (this was already true in most cases but could not be guaranteed because unrelated passes were allowed to treat the struct as a Java/PVL class)
- Enable getting the address of local and global variables (related: #1172)
  - Variables that have their address taken are automatically upgraded to pointer types
  - Currently no automatic permission annotations are generated for local variables so this may be confusing if you have a local variable that you need to declare access for in a loop invariant. (however in every case where this would occur you would not have been able to verify the program previously because of the address-of operator) Generating permission would likely be feasible
- Add a new pointer type that can never be null (also replaces some of the changes in #1239)
- Improve some of the names (which used to be `unknownN` in the resulting viper file to help with debugging
- Remove Viper field access from triggers where DerefPointer or PointerSubscript are the top-level expression (this is generally better for verification performance)
## PVL
- Add the `*` and `&` operators to PVL to work with pointer values (which could also already be used through the `pointer<T>` ADT)
- Add VerCors option `--contract-import-file` which allows importing method contracts and types from a PVL file which will replace the function signature/contract of the LLVM function with the same name (this will likely be removed later once Robert's specification stuff gets merged in but is necessary now to test the LLVM frontend)

# To do:
- [ ]  Update the wiki


pieter-bos:
Hey nice work, just a quick general comment while you work on this: I've realized that we're starting to do too much logic in the `Lang{_}ToCol` passes (i.e. the "unskippable" part), things should be in the transformation stage. The solution is to just adopt more nodes into the "core" of col: I think all reasonable language features should live there, so col-core is a ~superset of all our input languages. I don't mind duplicates (so by all means make a col equivalent of LLVM nodes so the initial translation is simple), but I'd encourage you to think about language features we should add into col.

superaxander:
> Hey nice work, just a quick general comment while you work on this: I've realized that we're starting to do too much logic in the `Lang{_}ToCol` passes (i.e. the "unskippable" part), things should be in the transformation stage. The solution is to just adopt more nodes into the "core" of col: I think all reasonable language features should live there, so col-core is a ~superset of all our input languages. I don't mind duplicates (so by all means make a col equivalent of LLVM nodes so the initial translation is simple), but I'd encourage you to think about language features we should add into col.

Yeah that makes sense I've so far not touched any of the transformation stuff so I don't quite know what goes in rewrite/, resolve/, etc. But indeed I already felt like some of the things I'm adding will be relevant more broadly.

superaxander:
My current plan is to remove the changes from #1172 such that this can be merged without affecting any other parts of the code. I had some issues with the new Ubuntu 24.02 image (which I'm using to get LLVM-17) but the last few runs seem to have worked instead of being suddenly cancelled.

superaxander:
Marking this as ready for review since I basically just want to add some tests now before merging this

src/col/vct/col/ast/declaration/global/ClassImpl.scala
```diff
@@ -1,18 +1,37 @@
...
+  def supports: Seq[Type[G]]
+  def intrinsicLockInvariant: Expr[G]
+
+  def classType(typeArgs: Seq[Type[G]]): TClass[G]
+
   def transSupportArrowsHelper(
       seen: Set[TClass[G]]
   ): Seq[(TClass[G], TClass[G])] = {
-    val t: TClass[G] = TClass(this.ref, typeArgs.map(v => TVar(v.ref)))
+    // TODO: Does this break things if we have a ByValueClass with supers?
```
bobismijnnaam: Why would it? Do you want to disallow plain assignment for byref classes?

src/col/vct/col/ast/declaration/singular/EndpointImpl.scala
```diff
@@ -12,7 +12,7 @@ trait EndpointImpl[G]
   override def layout(implicit ctx: Ctx): Doc =
     Group(Text("endpoint") <+> ctx.name(this) <+> "=" <+> init)
 
-  def t: TClass[G] = TClass(cls, typeArgs)
+  def t: TClass[G] = TByReferenceClass(cls, typeArgs)
```
bobismijnnaam: I guess this might be a bit of a code smell: while this is the right choice for now as it reflects what was already the case, I did not have in mind to restrict endpoints of choreographies to be either byref or byval. I only expect fields and methods.

Maybe we can discuss in-office if it's possible or not to leave it open for both here? I thought the whole point of this PR was to leave it open in rewriting code whether a class is byref or byval, but I'm sure I'm missing _a lot_ of (c++) context.

src/col/vct/col/ast/declaration/global/ClassImpl.scala
bobismijnnaam: The only thing I can come up with that's relevant is that in C, casting a MyStruct* to FirstMemberOfMyStruct* is well-defined. So in a way, a CStruct ast node does not have any supertypes, but the LangCToCol pass could compile it into a byvalue class, with the only supertype being equal to the first member of the struct. You'd have to account for that in assignment, though.

In addition, since you're targeting C, assigning different struct types makes no sense. So maybe struct types with supers should be disallowed? Then maybe byval classes should be separate from byref classes (not saying you have to do this, just thinking out loud)

src/col/vct/col/ast/lang/llvm/LLVMGlobalVariableImpl.scala
```diff
@@ -0,0 +1,10 @@
+package vct.col.ast.lang.llvm
+
+import vct.col.ast.LLVMGlobalVariable
+import vct.col.ast.ops.LLVMGlobalVariableOps
+import vct.col.print._
+
+trait LLVMGlobalVariableImpl[G] extends LLVMGlobalVariableOps[G] {
+  this: LLVMGlobalVariable[G] =>
+  // override def layout(implicit ctx: Ctx): Doc = ???
```
bobismijnnaam: Here and there there are some `layout` methods not implemented. Maybe for those where it makes sense you can implement a real one, and for others maybe put in a placeholder value that makes a bit more sense than "???"

src/col/vct/col/origin/Origin.scala
```diff
@@ -109,6 +109,9 @@ case class SourceName(name: String) extends NameStrategy {
     Some(SourceName.stringToName(name))
 }
 
+// Used to disambiguate whether to show a ByValueClass as a class or a struct
+case class TypeName(name: String) extends OriginContent
```
bobismijnnaam, line 113: If you focus on the C frontend, why distinguish between ByValueClass that require a "class" prefix? In addition, there's no PVL syntax, right? So might aswell settle on that ByValueClasses are structs, and that ByRef is class...?

src/col/vct/col/resolve/Resolve.scala
```diff
@@ -220,7 +229,12 @@ case object ResolveTypes {
           Spec.findModel(name, ctx)
             .getOrElse(throw NoSuchNameError("model", name, t))
         )
-      case t @ TClass(ref, _) =>
+      case t @ TByReferenceClass(ref, _) =>
+        ref.tryResolve(name =>
+          Spec.findClass(name, ctx)
+            .getOrElse(throw NoSuchNameError("class", name, t))
+        )
+      case t @ TByValueClass(ref, _) =>
```
bobismijnnaam, line 237: Just in case: if you make class/struct distinction, maybe you can change the "class" error message below to "struct" as well.

src/col/vct/col/resolve/Resolve.scala
```diff
@@ -1098,34 +1120,93 @@ case object ResolveReferences extends LazyLogging {
...
+                applicableContract.requires,
+                substitutedContract.requires,
+              ),
+              SplitAccountedPredicate(
+                applicableContract.ensures,
+                substitutedContract.ensures,
+              ),
+              applicableContract.contextEverywhere &*
+                substitutedContract.contextEverywhere,
+              applicableContract.signals ++ substitutedContract.signals,
```
bobismijnnaam, line 1178: Also a bit of a smell, not really a problem: there's not really a notion of "signals" in llvm, right? So the logic could be more precise here. There might be an exception for spec/ghost code, though that is currently also hypothetical (in the way that you'd need to guarantee that an exception thrown by ghost code is always caught, which we currently don't do)

src/col/vct/col/typerules/CoercionUtils.scala
```diff
@@ -300,6 +309,26 @@ case object CoercionUtils {
           case None => return None
         }
 
+      // TODO: Back and forth should not be needed...
```
bobismijnnaam: Maybe one of them should be an unsafe coercion? Similar to how a frac is a rat, but a rat is not necessarily a frac.

bobismijnnaam: Otherwise if you think it's  a bug when these coercions occur, it's also fine to disallow them.

src/col/vct/col/util/SubstituteReferences.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import vct.col.ast._
+import vct.col.origin.Origin
+import vct.col.ref.{DirectRef, Ref}
+import vct.col.rewrite.NonLatchingRewriter
+
+import scala.reflect.ClassTag
+
+/** Substitute all references in expressions, resulting AST can be used for
+  * analysis but not output since it doesn't contain the right declarations
```
bobismijnnaam: Maybe if you disable rewriting declarations in this rewriter, it might actually be safe? Since then it cannot create nodes that refer to declarations that were created later. A quick skim showed me that it was only used to rewrite expressions/types, so that restriction might work fine.

src/hre/hre/io/ChecksumReadableFile.scala
```diff
@@ -0,0 +1,55 @@
...
+        digest.digest(bytes).map(_.asInstanceOf[Int] & 0xff).map { b =>
+          if (b > 0xf) { Integer.toHexString(b) }
+          else { "0" + Integer.toHexString(b) }
+        }.mkString
+      )
+    } catch {
+      case _: NoSuchAlgorithmException =>
+        throw UnknownChecksumKind(checksumKind)
+    }
+    Files.newBufferedReader(file, StandardCharsets.UTF_8)
```
bobismijnnaam: Je leest hier de bytes 2 x toch? Kun je de bytearray niet als reader returnen? Je omzeilt de caching layer hier toch

src/rewrite/vct/rewrite/ClassToRef.scala
```diff
@@ -257,85 +366,338 @@ case class ClassToRef[Pre <: Generation]() extends Rewriter[Pre] {
...
+            if (cls.isInstanceOf[ByReferenceClass[Pre]]) {
+              byRefFieldSucc(field) =
+                new SilverField(dispatch(field.t))(field.o)
+              globalDeclarations.declare(byRefFieldSucc(field))
+            }
           case _ => throw ExtraNode
         }
-      case decl => rewriteDefault(decl)
+        cls match {
+          case cls: ByValueClass[Pre] =>
```
bobismijnnaam: Deze branch is beetje lang, kan dat niet in een aparte pass, of compacter gefactored? Als het moet dan moet t hoor maar ik dacht ik noem het even

src/rewrite/vct/rewrite/ConstantifyFinalFields.scala
```diff
@@ -80,12 +80,22 @@ case class ConstantifyFinalFields[Pre <: Generation]() extends Rewriter[Pre] {
         implicit val o: Origin = field.o
         if (isFinal(field)) {
           val `this` =
-            new Variable[Post](TClass(
-              succ(currentClass.top),
-              currentClass.top.typeArgs.map { v: Variable[Pre] =>
-                TVar(succ(v))
-              },
-            ))
+            currentClass.top match {
```
bobismijnnaam: De duplicatie beneden is ook wel een beetje een code smell, maar kan vast iet geunified worden met TClass...?

src/rewrite/vct/rewrite/DisambiguateLocation.scala
```diff
@@ -25,28 +34,43 @@ case object DisambiguateLocation extends RewriterBuilder {
...
 case class DisambiguateLocation[Pre <: Generation]() extends Rewriter[Pre] {
+  import DisambiguateLocation._
+
   def exprToLoc(expr: Expr[Pre], blame: Blame[PointerLocationError])(
       implicit o: Origin
   ): Location[Post] =
     expr match {
+      case expr if expr.t.asPointer.isDefined =>
+        PointerLocation(dispatch(expr))(blame)
+      case expr if expr.t.isInstanceOf[TByValueClass[Pre]] =>
```
bobismijnnaam: Mss "asValueClass.isDefined" en ook voor "asReferenceClass" toevoegen?

src/rewrite/vct/rewrite/EncodeAutoValue.scala
```diff
@@ -218,34 +218,36 @@ case class EncodeAutoValue[Pre <: Generation]() extends Rewriter[Pre] {
...
-                  Let(
-                    b,
-                    Old(v, None)(PanicBlame(
-                      "Old should always be valid in a postcondition"
-                    )),
-                    m,
-                  )
+            localHeapVariables.scope {
+              val top = conditionContext.pop()
+              val (b, v) =
```
bobismijnnaam, line 238: Dit is wel beetje sketchy :p

src/rewrite/vct/rewrite/EncodeAutoValue.scala
```diff
@@ -218,34 +218,36 @@ case class EncodeAutoValue[Pre <: Generation]() extends Rewriter[Pre] {
...
+              val m =
+                conditionContext.having((
+                  conditionContext.top._1,
+                  mMap,
+                  conditionContext.top._3,
+                  conditionContext.top._4,
+                )) { dispatch(main) }
+              if (mMap.isEmpty) { Let(b, v, m) }
+              else {
+                mMap.foreach(postM =>
```
bobismijnnaam, line 249: I'm probably just slow, maybe we can have a look at this code later

src/rewrite/vct/rewrite/EncodeForkJoin.scala
```diff
@@ -131,7 +131,13 @@ case class EncodeForkJoin[Pre <: Generation]() extends Rewriter[Pre] {
         implicit val o: Origin = e.o
         cls.decls.collectFirst { case run: RunMethod[Pre] => run } match {
           case Some(_) =>
-            val obj = new Variable[Post](TClass(succ(cls), Seq()))
+            val obj =
+              cls match {
```
bobismijnnaam: You had a "classType" helper for this, right...?

src/rewrite/vct/rewrite/EncodeIntrinsicLock.scala
```diff
@@ -143,7 +143,7 @@ case class EncodeIntrinsicLock[Pre <: Generation]() extends Rewriter[Pre] {
 
   override def dispatch(decl: Declaration[Pre]): Unit =
     decl match {
-      case cls: Class[Pre] =>
+      case cls: ByReferenceClass[Pre] =>
```
bobismijnnaam, line 147: Unrelated: should `intrinsicLockInvariant` for `ByValueClass` throw an exception or something? Or maybe should we remove it from the Class trait?

src/rewrite/vct/rewrite/EncodeResourceValues.scala
```diff
@@ -184,7 +184,11 @@ case class EncodeResourceValues[Pre <: Generation]()
             case ResourcePattern.HeapVariableLocation(_) => Nil
             case ResourcePattern.FieldLocation(f) =>
               nonGeneric(fieldOwner(f))
-              Seq(TClass(succ(fieldOwner(f)), Seq()))
+              Seq(fieldOwner(f) match {
+                case cls: ByReferenceClass[Pre] =>
```
bobismijnnaam: Helper?

src/rewrite/vct/rewrite/EncodeResourceValues.scala
```diff
@@ -196,8 +200,12 @@ case class EncodeResourceValues[Pre <: Generation]()
               ref.args.map(_.t).map(dispatch)
             case ResourcePattern.InstancePredicateLocation(ref) =>
               nonGeneric(predicateOwner(ref))
-              TClass[Post](succ(predicateOwner(ref)), Seq()) +:
-                ref.args.map(_.t).map(dispatch)
+              (predicateOwner(ref) match {
```
bobismijnnaam: Helper?

src/col/vct/col/ast/declaration/global/ClassImpl.scala
superaxander: I agree it's probably best to not support ByValueClasses with super types for now

src/hre/hre/io/ChecksumReadableFile.scala
superaxander: Yes I'm reading the bytes twice which was just an oversight but I'm not bypassing the caching layer since this overrides the method that is called by the caching layer

src/rewrite/vct/rewrite/ClassToRef.scala
superaxander: Yes there could be a separate ByRefClassToRef and ByValClassToRef or something like that. While I was writing this I didn't know how much these would diverge

src/rewrite/vct/rewrite/EncodeIntrinsicLock.scala
superaxander: For now I assumed it would be fine to just make the lock invariant a fixed `true`

src/rewrite/vct/rewrite/ExtractInlineQuantifierPatterns.scala
```diff
@@ -106,50 +106,56 @@ case class ExtractInlineQuantifierPatterns[Pre <: Generation]()
...
-            body = body,
-          )(f.o)
         }
 
       case f: Starall[Pre] =>
         variables.scope {
-          val (patternsHere, body) = patterns.collect {
-            // We only want to inline lets that are defined inside the quantifier
-            letBindings.having(ScopedStack()) { dispatch(f.body) }
+          localHeapVariables.scope {
```
bobismijnnaam, line 127: Nitpick: maybe the concept of variables and heap variables can be unified behind one scope/trait? They both have reasonable get/set implementations, and heap variables further extend that with an address. Just a thought, might not be worth the effort either?

src/rewrite/vct/rewrite/GenerateSingleOwnerPermissions.scala
```diff
@@ -82,7 +82,7 @@ case class GenerateSingleOwnerPermissions[Pre <: Generation](
           ),
         )
 
-      case cls: Class[Pre] if enabled =>
+      case cls: ByReferenceClass[Pre] if enabled =>
```
bobismijnnaam, line 85: As an exercise you might consider integrating heap variables into this pass, though maybe you already do so somewhere else?

src/rewrite/vct/rewrite/GenerateSingleOwnerPermissions.scala
```diff
@@ -287,20 +287,22 @@ case class GenerateSingleOwnerPermissions[Pre <: Generation](
...
-          foldStar(cls.collect { case f: InstanceField[Pre] =>
+      case t: TByReferenceClass[Pre]
+          if !generatingClasses.contains(t.cls.decl) =>
+        generatingClasses.having(t.cls.decl) {
+          foldStar(t.cls.decl.collect { case f: InstanceField[Pre] =>
             fieldTransitivePerm(e, f)(f.o)
           })
         }
-      case TClass(Ref(cls), _) =>
+      case t: TByReferenceClass[Pre] =>
```
bobismijnnaam: Why do you prefer type matching over pattern matching here?

src/rewrite/vct/rewrite/LowerLocalHeapVariables.scala
```diff
@@ -0,0 +1,96 @@
...
+  override def dispatch(program: Program[Pre]): Program[Post] = {
+    val dereferencedHeapLocals = program.collect {
+      case DerefPointer(hl @ HeapLocal(_)) => System.identityHashCode(hl)
+    }
+    val nakedHeapLocals = program.collect {
+      case hl @ HeapLocal(Ref(v))
+          if !dereferencedHeapLocals.contains(System.identityHashCode(hl)) =>
+        v
+    }
+    VerificationError.withContext(CurrentRewriteProgramContext(program)) {
```
bobismijnnaam, line 33: Can you not use program.rewrite here?

src/rewrite/vct/rewrite/LowerLocalHeapVariables.scala
```diff
@@ -0,0 +1,96 @@
...
+      }
+    }
+  }
+
+  override def dispatch(node: Statement[Pre]): Statement[Post] = {
+    implicit val o: Origin = node.o
+    node match {
+      // Same logic as CollectLocalDeclarations
+      case Scope(vars, impl) =>
+        val (newVars, newImpl) = variables.collect {
```
bobismijnnaam, line 51: Same here, scope.rewrite()?

bobismijnnaam: I don't see the difference, maybe it's even rewriteDefault in this case?

src/rewrite/vct/rewrite/MonomorphizeClass.scala
```diff
@@ -82,9 +82,16 @@ case class MonomorphizeClass[Pre <: Generation]()
...
             variables.scope {
-              allScopes.anyDeclare(
-                allScopes.anySucceedOnly(cls, cls.rewrite(typeArgs = Seq()))
-              )
+              localHeapVariables.scope {
+                allScopes.anyDeclare(allScopes.anySucceedOnly(
+                  cls,
+                  cls match {
+                    case cls: ByReferenceClass[Pre] =>
+                      cls.rewrite(typeArgs = Seq())
```
bobismijnnaam: You could make a classlike rewritehelper that rewrites only the common arguments?

src/rewrite/vct/rewrite/MonomorphizeClass.scala
```diff
@@ -130,14 +137,25 @@ case class MonomorphizeClass[Pre <: Generation]()
 
   override def dispatch(t: Type[Pre]): Type[Post] =
     (t, ctx.topOption) match {
-      case (TClass(Ref(cls), typeArgs), ctx) if typeArgs.nonEmpty =>
+      case (TByReferenceClass(Ref(cls), typeArgs), ctx) if typeArgs.nonEmpty =>
```
bobismijnnaam, line 135: Is it possible to match just on `: TClass[Pre]` here?

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+      }
+      case assign: Assign[Pre] =>
+        val target = inAssignment.having(()) { dispatch(assign.target) }
+        if (assign.target.t.isInstanceOf[TByValueClass[Pre]]) {
+          copyContext.having(InAssignmentStatement(assign)) {
+            assign.rewrite(target = target)
+          }
+        } else { assign.rewrite(target = target) }
+      case Instantiate(Ref(cls), out)
+          if cls.isInstanceOf[ByValueClass[Pre]] => {
```
bobismijnnaam: This can be in the pattern match, right?

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+              TrueSatisfiable,
+              classCreationMethods
+                .getOrElseUpdate(t, makeClassCreationMethod(t)).ref,
+            ),
+          )(AssignLocalOk),
+        ))
+      }
+      case assign: Assign[Pre] =>
+        val target = inAssignment.having(()) { dispatch(assign.target) }
+        if (assign.target.t.isInstanceOf[TByValueClass[Pre]]) {
```
bobismijnnaam: How about writing it like this?

```
case Assign(target, _) =>
  val newTarget = ...
  target.t match {
    case TByValueClass(_, _) =>
      copyContext.having(...) {
        assign.rewrite(target = newTarget)
      }
    case _ => assign.rewrite(target = newTarget)
  }
```

Possibly if we'd have a WithType(node, typeOf) extractor, maybe `:::`, you could check for the type in the first match. It's a bit of an unwritten rule but in most of the vercors codebase we try to avoid isInstanceOf/asInstanceOf.

You could go even further and have a separate assign statement for byvalue assigns, which makes sense as the semantics for regular assign and byvalue assign are quite different (byvalue-assign allocates!). But before that refactor it's probably best to take a few minutes to do a cost/benefit analysis.

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+        } else { assign.rewrite(target = target) }
+      case Instantiate(Ref(cls), out)
+          if cls.isInstanceOf[ByValueClass[Pre]] => {
+        // AssignLocalOk doesn't make too much sense since we don't know if out is a local
+        val t = TByValueClass[Pre](cls.ref, Seq())
+        Assign[Post](
+          dispatch(out),
+          procedureInvocation(
+            TrueSatisfiable,
+            classCreationMethods.getOrElseUpdate(t, makeClassCreationMethod(t))
```
bobismijnnaam: Usually successionmap fields we suffix with succ, so this would be "classCreationMethodsSucc". Sorry to be pedantic, but probably best to try to keep consistent for as long as possible now that Pieter is gone :)

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+      obj: Expr[Post],
+      t: TByValueClass[Pre],
+      blame: InstanceField[Pre] => Blame[InsufficientPermission],
+  ): Expr[Post] = {
+    implicit val o: Origin = obj.o
+    val ov = new Variable[Post](obj.t)(o.where(name = "original"))
+    val v = new Variable[Post](dispatch(t))(o.where(name = "copy"))
+    val children = t.cls.decl.decls.collect { case f: InstanceField[Pre] =>
+      f.t match {
+        case inner: TByValueClass[Pre] =>
```
bobismijnnaam: How about: `val myval = f.t match { case byval => copyClassValue...; case _ => deref ... } ; Assign(..., myVal)`?

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+        case _ => Perm(loc, perm)
+      }
+    })
+
+    foldStar(newFieldPerms)
+  }
+
+  override def dispatch(node: Expr[Pre]): Expr[Post] = {
+    implicit val o: Origin = node.o
+    node match {
```
bobismijnnaam: I think usually dispatch matches directly match on node types, if possible. Maybe in this case:

```
node match {
  case NewObject => ...
  case _ if inAssignment.nonEmpty() => ...
  ... all other cases ...
}
```
The only problem is that the implicit val o might be repeated here and there. But that's a bit of an unsolved problem, so better to just repeat that where necessary

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+    else
+      node match {
+        case Perm(ByValueClassLocation(e), p) =>
+          unwrapClassPerm(
+            dispatch(e),
+            dispatch(p),
+            e.t.asInstanceOf[TByValueClass[Pre]],
+          )
+        case Perm(pl @ PointerLocation(dhv @ DerefHeapVariable(Ref(v))), p)
+            if v.t.isInstanceOf[TNonNullPointer[Pre]] =>
```
bobismijnnaam: You don't use the outer expr here, so maybe put this in dispatch PointerLocation instead? Also, `pl` is unused I think.

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+          val t = v.t.asInstanceOf[TNonNullPointer[Pre]]
+          if (t.element.isInstanceOf[TByValueClass[Pre]]) {
+            val newV: Ref[Post, HeapVariable[Post]] = succ(v)
+            val newP = dispatch(p)
+            Perm(HeapVariableLocation(newV), newP) &* Perm(
```
bobismijnnaam: Probably you can inline the newV by: `Perm(HeapVariableLocation[Post](succ(v), ...)`. As a last resort, it is also acceptable to put a type annotation in the `succ[...]`

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+import vct.col.util.AstBuildHelpers._
+import vct.col.util.SuccessionMap
+import vct.result.VerificationError.{Unreachable, UserError}
+
+// TODO: Think of a better name
+case object PrepareByValueClass extends RewriterBuilder {
+  override def key: String = "prepareByValueClass"
+
+  override def desc: String =
+    "Initialise ByValueClasses when they are declared and copy them whenever they're read"
```
bobismijnnaam: "EncodeByValueClassCopySemantics"? CopySemantics/CopyBehaviour/CopyMethods/CopyInvocations/etc.

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+
+  private def rewriteInCopyContext(
+      dp: DerefPointer[Pre],
+      t: TByValueClass[Pre],
+  ): Expr[Post] = {
+    if (copyContext.isEmpty) {
+      // If we are in other kinds of expressions like if statements
+      return dp.rewriteDefault()
+    }
+    val clazz = t.cls.decl.asInstanceOf[ByValueClass[Pre]]
```
bobismijnnaam: Maybe you can use "cls" here instead? Bit of a bikeshed maybe but I've always disliked the weird variations on class/this...

src/rewrite/vct/rewrite/PrepareByValueClass.scala
```diff
@@ -0,0 +1,333 @@
...
+        case _ => node.rewriteDefault()
+      }
+  }
+
+  private def rewriteInCopyContext(
+      dp: DerefPointer[Pre],
+      t: TByValueClass[Pre],
+  ): Expr[Post] = {
+    if (copyContext.isEmpty) {
+      // If we are in other kinds of expressions like if statements
```
bobismijnnaam: Sorry, nitpicky idea: preload the copyContext with a top-level (maybe NoCopy?) element so top is always safe. Then you can get rid of the early return :)

src/rewrite/vct/rewrite/ResolveExpressionSideEffects.scala
```diff
@@ -680,7 +659,19 @@ case class ResolveExpressionSideEffects[Pre <: Generation]()
             givenMap,
             yields,
           ) =>
-        val typ = TClass[Post](succ(cons.cls.decl), classTypeArgs.map(dispatch))
+        val typ =
```
bobismijnnaam: Can you rewrite this using t.inv.rewrite?

src/rewrite/vct/rewrite/ResolveExpressionSideEffects.scala
```diff
@@ -695,12 +686,25 @@ case class ResolveExpressionSideEffects[Pre <: Generation]()
...
         )
-        stored(res.get(SideEffectOrigin), TClass(cons.cls, classTypeArgs))
+        stored(
+          res.get(SideEffectOrigin),
+          cons.cls.decl.classType(classTypeArgs),
+        )
       case NewObject(Ref(cls)) =>
-        val res = new Variable[Post](TClass(succ(cls), Seq()))(ResultVar)
+        val res =
+          cls match {
```
bobismijnnaam: Also use the type rewrite trick here?

src/rewrite/vct/rewrite/SimplifyNestedQuantifiers.scala
```diff
@@ -228,6 +228,18 @@ case class SimplifyNestedQuantifiers[Pre <: Generation]()
     )(contract.blame)(contract.o)
   }
 
+  private def hasTriggers(e: Binder[Pre]): Boolean =
+    e match {
+      case Forall(_, triggers, body) =>
```
bobismijnnaam, line 233: Honestly maybe we could argue for a "Quantifier" trait that extends binder to generalize forall/exists... Not sure if there would be any use elsewhere in vercors tho

src/rewrite/vct/rewrite/VariableToPointer.scala
```diff
@@ -0,0 +1,219 @@
...
+      case Local(Ref(v)) if addressedSet.contains(v) =>
+        DerefPointer(Local[Post](variableMap.ref(v)))(PanicBlame(
+          "Should always be accessible"
+        ))
+      case deref @ Deref(obj, Ref(f)) if addressedSet.contains(f) =>
+        DerefPointer(Deref[Post](dispatch(obj), fieldMap.ref(f))(deref.blame))(
+          PanicBlame("Should always be accessible")
+        )
+      case newObject @ NewObject(Ref(cls))
+          if cls.isInstanceOf[ByValueClass[Pre]] =>
```
bobismijnnaam: You can put this in a pattern match in the type, right?

src/rewrite/vct/rewrite/lang/LangLLVMToCol.scala
```diff
@@ -2,57 +2,305 @@ package vct.rewrite.lang
...
+
+    // TODO: This sorting is non-stable which might cause nondeterministic bugs if there's something wrong with moreSpecific
+    def findMostSpecific(
+        types: mutable.ArrayBuffer[Type[Pre]]
+    ): Option[Type[Pre]] = {
+      types.map(Some(_)).reduce[Option[Type[Pre]]] { (a, b) =>
+        (a, b) match {
+          case (None, _) | (_, None) => None
+          case (Some(a), Some(b))
+              if a == b || rw.dispatch(a) == rw.dispatch(b) ||
```
bobismijnnaam, line 100: Comparing things in `Post` is highly dubious! Might trigger comparison of some uninitialized refs. I know sycl used to do this every once in a while and it's a pain.

Is it maybe possible to make a more explicit ordering for a fragment of types? As this is for llvm, you only need to do this for say ints, structs, and the spec types (seq/map etc.). And for the spec types you can maybe just say to not let them interleave with llvm types maybe, to simplify things? Relying on the behaviour of another rewriter is fragile. It's also fine if your ordering is very limited and needs to be extended every once in a while. What about only using `moreSpecific`, and maybe making `moreSpecific` more complete for the identity cases?

src/rewrite/vct/rewrite/lang/LangLLVMToCol.scala
```diff
@@ -2,57 +2,305 @@ package vct.rewrite.lang
...
+            new LLVMGlobalVariable[Pre](
+              typeGuesses(v).currentType,
+              v.value,
+              v.constant,
+            )(v.o),
+          )
+      }
+      if (subMap.isEmpty) { value }
+      else {
+        // TODO: Support multiple guesses?
```
bobismijnnaam: Only if there's an actual use case, I don't imagine any sane compiler does this (I'm probably wrong! :D)

src/rewrite/vct/rewrite/lang/LangSpecificToCol.scala
```diff
@@ -317,17 +336,17 @@ case class LangSpecificToCol[Pre <: Generation](
...
-        c.getBaseType(expr.t) match {
-          case structType: CTStruct[Pre] =>
-            c.unwrapStructPerm(
-              dispatch(a).asInstanceOf[AmbiguousLocation[Post]],
-              perm,
-              structType,
-              e.o,
-            )
-        }
+//      case Perm(a @ AmbiguousLocation(expr), perm)
```
bobismijnnaam: Maybe you can just delete this? Otherwise please also comment why it is important to keep

src/rewrite/vct/rewrite/lang/LangTypesToCol.scala
```diff
@@ -88,7 +88,10 @@ case class LangTypesToCol[Pre <: Generation]() extends Rewriter[Pre] {
       case t @ PVLNamedType(_, typeArgs) =>
         t.ref.get match {
           case spec: SpecTypeNameTarget[Pre] => specType(spec, typeArgs)
-          case RefClass(decl) => TClass(succ(decl), typeArgs.map(dispatch))
+          case RefClass(decl: ByReferenceClass[Pre]) =>
+            TByReferenceClass(succ[Class[Post]](decl), typeArgs.map(dispatch))
```
bobismijnnaam: Probably you can put the `[Post]` on TByRef, and then you don't need it on succ

src/rewrite/vct/rewrite/lang/NoSupportSelfLoop.scala
```diff
@@ -13,7 +13,15 @@ case object NoSupportSelfLoop extends RewriterBuilder {
 case class NoSupportSelfLoop[Pre <: Generation]() extends Rewriter[Pre] {
   override def dispatch(decl: Declaration[Pre]): Unit =
     decl match {
-      case cls: Class[Pre] =>
+      case cls: ByReferenceClass[Pre] =>
+        globalDeclarations.succeed(
+          cls,
+          cls.rewrite(supports =
```
bobismijnnaam, line 18: Here again maybe you can make a `.rewrite` helper for ByRef/ByVal classes

bobismijnnaam (commented):
There you go! Good work!

src/rewrite/vct/rewrite/PrepareByValueClass.scala
superaxander: What do you mean by dispatch PointerLocation? Also `pl` is used in `pl.blame`

src/rewrite/vct/rewrite/PrepareByValueClass.scala
superaxander: `newV ` is a variable bescause it's used twice (and it needed the type annotation so I thought I'd kill two birds with one stone)

src/rewrite/vct/rewrite/lang/LangLLVMToCol.scala
superaxander: Yeah I knew this was dubious when I wrote it. I do it because I'm mixing LLVM and PVL in places and their types will only match after the LangSpecificToCol pass has finished. This was mostly a cop-out for not having to write out every LLVM type that might match a PVL type and vice-versa.

PR #1167, opened 2024-03-05
Add challenge 3 from the verifyThis 2019 challenge

sakehl:
Add example for solving challenge 3 from the verifyThis 2019 challenge.

Also add the possibility to test with a specific flag. Needed since otherwise the challenge does not verify.


bobismijnnaam:
Preemptive apologies for not understanding your use case. I have a few tests that use flags already, see here: https://github.com/utwente-fmt/vercors/blob/72735981dd3ab612f93aa918d6b152f80dee9d7e/test/main/vct/test/integration/examples/VeyMontExamplesSpec.scala#L12 Does that work for you? If that is somehow not what you need please ignore my comment :)

PR #1170, opened 2024-03-21
Bool/Int type coercion

ArmborstL:
Before merging this PR, please check the following:

- [x] I made sure the wiki is updated in accordance with the changes in this PR. For example, syntax changes, semantics changes, VerCors flags changes, etc.

# PR description
Add type coercion between `int` and `bool` for C. Also coerce pointers to `bool`.

In C, integers can act as booleans (e.g. in `if(myInt)`), and booleans are just integers internally and can be passed as such, e.g. in 
```c
void check(int i) {}
void main() {
  check(0==0);
}
```
Therefore, implicit casting between those types should work.

Pointers are basically integers, so the same applies for them. E.g. after `p=malloc(...)`, we might write `if(!p)return;`, treating the pointer as a boolean. This PR also covers that case.


src/col/vct/col/ast/family/coercion/CoercionImpl.scala
ArmborstL, line 1: @sakehl I'm not sure what isCPromoting does. Should I add one of the coercions as a case there? In principle, I want all coercions to work in C code, so I'm reluctant to add e.g. CoerceCIntBool as `false` here, but it is not injective and thus not promoting...
Also, is it correct that some cases, like CoerceJoinUnion, call the regular `isPromoting` instead of the special C version `isCPromoting`? or is that just a copy-paste error?

sakehl: What I wanted to capture here is that float/double can only be `demoted` to int (and double to float) in:
* an assignment
* an argument of a function call
I based this mostly on this [reference](https://www.oreilly.com/library/view/c-in-a/0596006977/ch04.html) and [here](https://en.cppreference.com/w/c/language/conversion)

This is called demotion, and is only valid in those cases. More specifically it is not valid with arithmetic conversions. (so `5 % 4.0` is not a valid statement, since 4.0 cannot be demoted to an int here, and % does not work on floating point numbers).

Anyway it seems that boolean is a bit harder. When looking specifically at if-statements and logical and/or statements of C:
https://en.cppreference.com/w/c/language/operator_logical

They work for `an expression of any scalar type`. Which is of course not how we defined this internally in VerCors.

I think we can then best encode this to always allow `demoting` pointers and integers to booleans.

So in that case, I'd say that we put:
```
case CoerceCIntBool() => false
    case CoercePointerBool() => false
  }
```
both to true.

(.... what is C a messy language)

> Also, is it correct that some cases, like CoerceJoinUnion, call the regular isPromoting instead of the special C version isCPromoting? or is that just a copy-paste error?

This is definitely a copy paste error, thanks for spotting it.

src/col/vct/col/typerules/CoercionUtils.scala
```diff
@@ -90,6 +90,11 @@ case object CoercionUtils {
       case (TFraction(), TRational()) => CoercionSequence(Seq(CoerceFracZFrac(), CoerceZFracRat()))
       case (TZFraction(), TRational()) => CoerceZFracRat()
       case (source: FloatType[G], TRational()) => CoerceFloatRat(source)
+      case (TBool(), TCInt()) => CoerceBoolCInt()
+      case (TCInt(), TBool()) => CoerceCIntBool()
+      case (TCInt(), TResource()) => CoerceCIntBool()
```
sakehl, line 95: Maybe:
`CoercionSequence(Seq( CoerceCIntBool(), CoerceBoolResource()))`

test/main/vct/test/integration/examples/TechnicalSpec.scala
```diff
@@ -434,6 +434,8 @@ vercors should verify using silicon in "example using string primitive" pvl
 
   vercors should verify using anyBackend example "technical/labeledEmbeddedStatement.c"
 
+  vercors should verify using anyBackend example "technical/intBoolCoercion.c"
```
sakehl, line 437: Does this actually pass? I would not expect it without the changes in CPromoting

ArmborstL: Yes, it does. Note that `isCPromoting` right now does not have cases for CoerceCIntBool and CoerceBoolCInt, so the default `true` applies. The two `false` cases you mention in your comment above are in `isPromoting`, not in `isCPromoting`.

src/col/vct/col/ast/family/coercion/CoercionImpl.scala
ArmborstL: > I think we can then best encode this to always allow demoting pointers and integers to booleans.

Ok, then I leave it as-is and let the default case handle it.
> This is definitely a copy paste error

Ok, I'll fix it :)

superaxander:
Implemented in #1405 and #1406

Issue #1165, opened 2024-03-01
Parallel Block encoder, saving ranges as variables


sakehl:
So in the parallel block encoding ParBlockEncoder:

https://github.com/utwente-fmt/vercors/blob/72735981dd3ab612f93aa918d6b152f80dee9d7e/src/rewrite/vct/rewrite/ParBlockEncoder.scala#L260-L265

I've added this code such that if something like
```
void main(int N){
  par(tid=0..N)
    requires E1
    ensures E2
 {
    ...
  }
}
```
will be rewritten such that rangeValue for high, is just saved as `N` (_wanted case_) instead of `hi_tid=N`. (_fallback case_)

This is important, since E1 and E2 get automatically quantified over the range of tid.

In the case we have just saved N, we get:
`(forall int tid=0..N; E1);`
instead of
`(forall int tid=0..hi_tid; E1);`

And because we have no access hi_tid, and cannot add annotation about hi_tid.

And if we ever want to rewrite quantifiers or reason about things, this becomes hard/impossible. This is especially essential for any GPU kernels, since they each have 6 parallel blocks (3 thread-blocks and 3 local-threads).

Anyway, the approach now, is working, but is brittle. Since it is a a conservative safe analysis, but if it comes across a node it does not know, it goes to the fallback case. 

I was thinking, we could also analyse any referred variables.

If the variable referenced is of primitive type or an ADT type, and the variable is local/a parameter, it cannot be altered in a parallel block.
(This is correct right?)

If we can conclude this, we can end up in the _wanted_ case.

I posted this here, mainly to see if this is a correct assumption or if you might have other ideas. Especially @pieter-bos if you have an opinion on this.

If you think this would work, I'll implement this.

pieter-bos:
I'm struggling to understand a bit under what circumstances you need to explicitly reason about the upper bound (i.e. `hi_tid`) - it's a constant with respect to the parallel block so you should have local inference about its value. In the case that the bound value is heap-dependent you may have problems if you are also using `--no-infer-heap-context-into-frame`. Or maybe you are talking about your rewriting analysis, but there it should be fairly trivial to just internalize that `hi_tid == N`, or?

Anyway, I don't have strong objections to do something like this, but please be careful about well-formedness constraints. Even when you can ascertain that an evaluation would not change, it's possible we can't conclude from the local thread contract that the expression is still well-formed (think division by zero, sequence subscript, that sort of thing).

PR #1172, opened 2024-03-25
Allow the use of pointers to things that are not a reference type

superaxander:
Currently you can only take the address of a struct in the C and C++ frontends. By converting variables to pointers whenever you take their address you can do the same for non-reference types.

This is a draft PR because I'm still working on it. While it is currently at a state where I'm not breaking anything and simple (nonsense) examples like this verify:
```c
struct A {
    int a;
    int b;
    int *c;
};


//@ ensures \result == 10;
int main() {
    struct A a;

    a.a = 2;

    struct A* b = &a;
    b->a = 5;
    b->b = 5;
    b->c = &a.a;
    return a.a + a.b;
}
```
I am unsure where the limits are here. Additionally I still need to fix the locations because you cannot write the necessary loop invariants to preserve access to your local which was transformed to a pointer.

### DIFF
--- a/src/main/vct/main/stages/Transformation.scala
+++ b/src/main/vct/main/stages/Transformation.scala
@@ -209,9 +209,9 @@ case class SilverTransformation
     EncodeChar,
 
     CollectLocalDeclarations, // all decls in Scope
+    VariableToPointer, // should happen before ParBlockEncoder so it can distinguish between variables which can and can't altered in a parallel block
     DesugarPermissionOperators, // no PointsTo, \pointer, etc.
     ReadToValue, // resolve wildcard into fractional permission
-    VariableToPointer,
     TrivialAddrOf,
     DesugarCoalescingOperators, // no ?.
     PinCollectionTypes, // no anonymous sequences, sets, etc.
--- a/src/main/vct/main/stages/Transformation.scala
+++ b/src/main/vct/main/stages/Transformation.scala
@@ -23,7 +23,7 @@ import vct.options.types.{Backend, PathOrStd}
 import vct.resources.Resources
 import vct.result.VerificationError.SystemError
 import vct.rewrite.adt.ImportSetCompat
-import vct.rewrite.{EncodeRange, EncodeResourceValues, ExplicitResourceValues, HeapVariableToRef, SmtlibToProverTypes}
+import vct.rewrite.{EncodeRange, EncodeResourceValues, ExplicitResourceValues, HeapVariableToRef, SmtlibToProverTypes, VariableToPointer}
 import vct.rewrite.lang.ReplaceSYCLTypes
 import vct.rewrite.veymont.{DeduplicateSeqGuards, EncodeSeqBranchUnanimity, EncodeSeqProg, EncodeUnpointedGuard, GenerateSeqProgPermissions, SplitSeqGuards}
 
@@ -211,6 +211,7 @@ case class SilverTransformation
     CollectLocalDeclarations, // all decls in Scope
     DesugarPermissionOperators, // no PointsTo, \pointer, etc.
     ReadToValue, // resolve wildcard into fractional permission
+    VariableToPointer,
     TrivialAddrOf,
     DesugarCoalescingOperators, // no ?.
     PinCollectionTypes, // no anonymous sequences, sets, etc.
--- a/src/rewrite/vct/rewrite/VariableToPointer.scala
+++ b/src/rewrite/vct/rewrite/VariableToPointer.scala
@@ -56,8 +56,17 @@ case class VariableToPointer[Pre <: Generation]() extends Rewriter[Pre] {
       } ++ Seq(dispatch(s.body))))
       case i@Instantiate(cls, out) =>
         Block(Seq(i.rewriteDefault()) ++ cls.decl.declarations.flatMap {
-          case f: InstanceField[Pre] if addressedSet.contains(f) =>
-            Seq(Assign(Deref[Post](dispatch(out), fieldMap.ref(f))(PanicBlame("Initialisation should always succeed")), NewPointerArray(fieldMap(f).t.asPointer.get.element, const(1))(PanicBlame("Size is > 0")))(PanicBlame("Initialisation should always succeed")))
+          case f: InstanceField[Pre]  =>
+            if (f.t.asClass.isDefined) {
+              Seq(
+                Assign(Deref[Post](dispatch(out), fieldMap.ref(f))(PanicBlame("Initialisation should always succeed")), NewPointerArray(fieldMap(f).t.asPointer.get.element, const(1))(PanicBlame("Size is > 0")))(PanicBlame("Initialisation should always succeed")),
+                Assign(PointerSubscript(Deref[Post](dispatch(out), fieldMap.ref(f))(PanicBlame("Initialisation should always succeed")), const[Post](0))(PanicBlame("Size is > 0")), dispatch(NewObject[Pre](f.t.asClass.get.cls)))(PanicBlame("Initialisation should always succeed"))
+              )
+            } else if (addressedSet.contains(f)) {
+              Seq(Assign(Deref[Post](dispatch(out), fieldMap.ref(f))(PanicBlame("Initialisation should always succeed")), NewPointerArray(fieldMap(f).t.asPointer.get.element, const(1))(PanicBlame("Size is > 0")))(PanicBlame("Initialisation should always succeed")))
+            } else {
+              Seq()
+            }
           case _ => Seq()
         })
       case other => other.rewriteDefault()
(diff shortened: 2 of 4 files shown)


sakehl:
Ow interesting, I've thought a lot about this as well.

structs and fields of structs should work easily indeed, since they are translated to `refs` in viper already.

More interesting are cases like:

```
 //@ context x!= NULL && \pointer_length(x) == 1 && Perm(x, write);
 //@ ensures *x == \old(*x)+1;
void f(int* x){
  x++;
}

void main(){
  int x = 5;
  f(&x);
  //@ assert x == 6;
}
```

What I've been thinking is maybe adding additional annotations when declaring a var, such that we actually consider it a memory location for which we have can specify permissions, since it seems like an overhead of the encoding to encode every basic type as a memory location.

So maybe something like:
```
/*@ ref @*/ int x;
```
or something.

Probably @pieter-bos has some ideas here too.

sakehl:
Looking at the code, you make a variable a pointer, whenever address of is used of that variable. Which is nice.

I wonder how that works with parallel blocks and for loops. Do you then need to also add for those variables permissions, that they are not NULL and have pointer_length of 1?

Anyway, nice work!

superaxander:
Yeah so what I've done for now is simply detect which variables and fields get their address taken and "upgrade" those to pointers automatically. I'm not sure if there are cases where this could not be detected and such an annotation would be needed. I was trying some stuff out in VeriFast and it seems to take a similar approach of implicitly upgrading variables to pointers.

For example in this program you can use `==` to reason about the value of `a` in the loop invariant:
```c
int main()
    //@ requires true;
    //@ ensures result == 10;
{
    int a = 15;    
    int x = 0;
    while (x < 1)
    	//@ invariant x >= 0 && x <= 1 &*& a == ((x == 1) ? 10 : 15);
    {
    	a = 10;
    	x += 1;
    }
    return a;
}
```

But if I add a line `ìnt *b = &a` anywhere in this function then I have to rewrite the invariant to look like this:
```c
x >= 0 && x <= 1 &*& a |-> ((x == 1) ? 10 : 15);
```
And removing the `int *b = &a` will require you to also switch back to `==`.

superaxander:
> Looking at the code, you make a variable a pointer, whenever address of is used of that variable. Which is nice.
> 
> I wonder how that works with parallel blocks and for loops. Do you then need to also add for those variables permissions, that they are not NULL and have pointer_length of 1?
> 
> Anyway, nice work!

Yes that is the reason why this a draft PR currently. Things break when you have loops and currently there is no way to write the necessary loop invariant to say that you keep access to the pointer after the loop, but that is probably just a mistake I made in the location transformation.

src/main/vct/main/stages/Transformation.scala
```diff
@@ -210,6 +210,7 @@ case class SilverTransformation
     CollectLocalDeclarations, // all decls in Scope
     DesugarPermissionOperators, // no PointsTo, \pointer, etc.
     ReadToValue, // resolve wildcard into fractional permission
+    VariableToPointer,
```
sakehl, line 213: I think it would be best to keep this after `ParBlockEncoder` (which it is), because of this: https://github.com/utwente-fmt/vercors/issues/1165 . Maybe add a comment saying this.

src/rewrite/vct/rewrite/VariableToPointer.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  var stageTwo = false
+  val addressedSet: mutable.Set[Node[Pre]] = new mutable.HashSet[Node[Pre]]()
+  val heapVariableMap: SuccessionMap[HeapVariable[Pre], HeapVariable[Post]] = SuccessionMap()
+  val variableMap: SuccessionMap[Variable[Pre], Variable[Post]] = SuccessionMap()
+  val fieldMap: SuccessionMap[InstanceField[Pre], InstanceField[Post]] = SuccessionMap()
+
+  override def dispatch(program: Program[Pre]): Program[Rewritten[Pre]] = {
+    super.dispatch(program)
+    stageTwo = true
```
sakehl, line 37: Instead of two stages, maybe you can do:

`program.collect({case a :AddrOf[Pre] =>a})`
to collect all adrr offs and process them separately.

superaxander: Ah I didn't know that existed

superaxander:
I moved the pass to be before DesugarPermissionOperators now so that you can use \pointer and \pointer_index with variables which were automatically turned into pointers. I also changed those two to also imply permission to the fields of a class if you have a pointer to a class. Not sure if that is actually a great idea but it doesn't seem to break any existing tests. (the only test which didn't pass on my machine was `examples/publications/2020/permutations/oddPhase.pvl` but it doesn't seem to be doing anything which I could've affected) If I revert this change it would probably be good to add an alternative because otherwise specifying permissions for pointers to structs which are local variables becomes very unwieldy. In fact it is currently not possible to specify these permissions without using \pointer in some more complicated cases.

I also changed the TrivialAddrOf so that you can now check if a pointer is equal to the address of a variable. However I probably have to get rid of this and replace it with something that makes more sense because the current translation of `ptr_a == &a` is `*ptr_a == a` which would imply that the values being equal implies the pointers are equal which is obviously not the case. A better translation would be `ptr_a == a`. (where `a` has been turned into a pointer to the actual value)

However, right now I am not touching variables which were already stored as references (i.e. classes) since if I start transforming those to pointers then a bunch of existing examples would require additional annotations. I am still trying to think of what would be the best way to preserve the existing behaviour while also allowing you to properly reason about the addresses of these reference types. 

This now allows us to verify a program like this:
```c
struct A {
    int a;
    int b;
};


//@ ensures \result == 10;
int main() {
    struct A a;

    a.a = 2;

    struct A* b = &a;

    //@ loop_invariant \pointer(b, 1, write);
    //@ loop_invariant b->a == 5 ==> b->b == 5;
    //@ loop_invariant b == &a;
    while (b->a != 5) {
        b->a = 5;
        b->b = 5;
    }

    return a.a + a.b;
}
```

superaxander:
Right now the following program can be verified:

```c
struct A {
    int a;
    int b;
};
struct B {
    struct A a;
};
//@ ensures \result == 10;
int main() {
    struct B b;

    b.a.a = 2;

    struct A* a = &b.a;

    int *c = &a->a;
    int *d = &a->b;

    //@ loop_invariant Perm(a, write);
    //@ loop_invariant Perm(&a->a, write);
    //@ loop_invariant Perm(&a->b, write);
    //@ loop_invariant c == &a->a;
    //@ loop_invariant d == &a->b;
    //@ loop_invariant \pointer(&a->a, 1, write);
    //@ loop_invariant \pointer(&a->b, 1, write);
    //@ loop_invariant *c == 5 ==> *d == 5;
    while (*c != 5) {
        *c = 5;
        *d = 5;
    }

    return a->a + a->b;
}
```

To get this to work I had initialise the inner struct whenever the outer struct is initialised since otherwise you don't have permission to access the inner struct. (I guess that was just an open issue with structs currently?)

I also added a bit of a hack

https://github.com/utwente-fmt/vercors/blob/5386e5005cf236ba7147e7c80dc24f96c514374c/src/rewrite/vct/rewrite/VariableToPointer.scala#L110-L112

Which gets rid of an initial AddrOf in a PointerLocation so that you can actually refer to the implicit pointer and say something about its permissions.

superaxander:
An additional note about this:
>To get this to work I had initialise the inner struct whenever the outer struct is initialised since otherwise you don't have permission to access the inner struct. (I guess that was just an open issue with structs currently?)

I currently implemented this by simply initialising any classes contained in the class when we encounter a NewObject/Instantiate statement/expression. That would be a problem if it happened to Java classes since they should of course have null fields by default. However, the `new` expressions in Java and PVL are translated to ConstructorInvocations so we do not have this problem. But maybe to avoid confusion this default initialising behaviour should be implemented in LangCToCol and LangCPPToCol instead.

sakehl:
> An additional note about this:
> 
> > To get this to work I had initialise the inner struct whenever the outer struct is initialised since otherwise you don't have permission to access the inner struct. (I guess that was just an open issue with structs currently?)
> 
> I currently implemented this by simply initialising any classes contained in the class when we encounter a NewObject/Instantiate statement/expression. That would be a problem if it happened to Java classes since they should of course have null fields by default. However, the `new` expressions in Java and PVL are translated to ConstructorInvocations so we do not have this problem. But maybe to avoid confusion this default initialising behaviour should be implemented in LangCToCol and LangCPPToCol instead.

Tbh: what should really happen is that we give Struct it's own internal type and declerations.

I `hacked` in struct to be similar to a class, just to get something working. But really they are different beast, where structs are really 'values' and classes are references. 

So we should have a TStruct as Type, and Struct[G] as decleration together with StructDeclaration etc in Node.scala

superaxander:
Superseded by #1159

PR #1182, opened 2024-03-29
Create an docker image based on ubuntu as well

sakehl:


pieter-bos:
I've just made ubuntu the default base. I think this is more convenient for anyone needing to build an artifact in the future, and it's not worth it to fiddle with alpine compatibility issues for the size decrease. See this commit: <https://github.com/utwente-fmt/vercors/commit/1c19477f88d4f266234ca40f5770671fcb074a2d>

Issue #1195, opened 2024-05-07
Slow VerCors passes for simple program


sakehl:
I wanted to share the following programs, which I investigated for run times.
These are the verification times:

data/quant_s1.pvl 8.662854194641113 success
data/quant_s2.pvl 7.364089488983154 success
data/quant_s3.pvl 7.336841583251953 success
data/quant_s4.pvl 7.506519317626953 success
data/quant_s5.pvl 7.664606809616089 success
data/quant_s6.pvl 7.704848051071167 success
data/quant_s7.pvl 7.864739894866943 success
data/quant_s8.pvl 7.981231689453125 success
data/quant_s9.pvl 8.1972496509552 success
data/quant_s10.pvl 8.15115475654602 success
data/quant_s11.pvl 8.232187032699585 success
data/quant_s12.pvl 9.20827865600586 success
data/quant_s13.pvl 9.747354984283447 success
data/quant_s14.pvl 10.299208879470825 success
data/quant_s15.pvl 11.907938718795776 success
data/quant_s16.pvl 15.329325914382935 success
data/quant_s17.pvl 18.94451141357422 success
data/quant_s18.pvl 27.495386838912964 success
data/quant_s19.pvl 45.91024327278137 success

Investigating the slow down, it seems that nearly _all_ the time is spend with VerCors. Doing all the passes. Actually verification takes maybe a second or 2 at most.

[quant_s.zip](https://github.com/utwente-fmt/vercors/files/15238012/quant_s.zip)

This is good example program:

```java

  context_everywhere n > 0;
  context_everywhere x0 != null ** x0.length == n ** (\forall* int i=0..n; Perm(x0[i], write));
  context x1 != None && |x1.get| == n;
  context x2 != None && |x2.get| == n;
  context x3 != None && |x3.get| == n;
  context x4 != None && |x4.get| == n;
  context x5 != None && |x5.get| == n;
  context x6 != None && |x6.get| == n;
  context x7 != None && |x7.get| == n;
  context x8 != None && |x8.get| == n;
  context x9 != None && |x9.get| == n;
  context x10 != None && |x10.get| == n;
  context x11 != None && |x11.get| == n;
  context x12 != None && |x12.get| == n;
  context x13 != None && |x13.get| == n;
  context x14 != None && |x14.get| == n;
  context x15 != None && |x15.get| == n;
  context x16 != None && |x16.get| == n;
  context x17 != None && |x17.get| == n;
[...]

PR #1197, opened 2024-05-09
Small fix

sakehl:
Some small fixes I needed:
* Increases stack size to 128m: I was running into stack overflow errors again. But if you rather have dev not have such a big stack size, I'll keep these changes local.
* Added almost all math.h functions: All trigonometric are imported as abstract functions, with some properties attached. Mainly making sure code that uses these does not immediately break down.
* I tried to consistently remove `CPrimitiveType` from some places (ran into bugs otherwise not verifying, like inlined float functions which did not work)
* I want to use predicates in triggers, these changes allow this (if it sees a predicateApply in triggers, it does not use the fractional permission, but does use the predicate).
* Made some types lazy, fixing #1195 : Although in that issue there are still relatively slow passes for simplify we might want to look at before closing it completely.


pieter-bos:
If you have a reproducer for an example needing so much stack memory I'd appreciate it :)

sakehl:
So I have this file:

After parsing takes 10minutes, it gets to the verification. On which it stack overflows eventually after approx 5 minutes.

[SolveDirectionHalide-CB-P.zip](https://github.com/utwente-fmt/vercors/files/15260970/SolveDirectionHalide-CB-P.zip)

src/viper/viper/api/transform/ColToSilver.scala
```diff
@@ -39,6 +39,8 @@ case class ColToSilver(program: col.Program[_]) {
   val currentMapGet: ScopedStack[col.MapGet[_]] = ScopedStack()
   val currentDividingExpr: ScopedStack[col.DividingExpr[_]] = ScopedStack()
 
+  var inTriggers = false
```
pieter-bos: please use a `ScopedStack` for this, I plan to enable parallelism within rewriters at some point.

pieter-bos:
I've added a couple flags to diagnose potential causes of parsing slowness [here](https://github.com/utwente-fmt/vercors/wiki/Parser-Analysis)

sakehl:
> I've added a couple flags to diagnose potential causes of parsing slowness [here](https://github.com/utwente-fmt/vercors/wiki/Parser-Analysis)


> So I have this file:
> 
> After parsing takes 10minutes, it gets to the verification. On which it stack overflows eventually after approx 5 minutes.
> 
> [SolveDirectionHalide-CB-P.zip](https://github.com/utwente-fmt/vercors/files/15260970/SolveDirectionHalide-CB-P.zip)


Ah, excuse me, my terminology was off. I did not mean "parsing", I meant that the complete process of parsing + transformations takes about 10 minutes. Especially frtom the "adtNull" transformation and onward it takes a lot of time.

The actually parsing is like a few seconds.

pieter-bos:
No worries, parsing performance is a recurring problem anyway. Seconds is already 5-6 orders of magnitude slower than it should be, really.

sakehl:
ok should be ok now.

Issue #502, opened 2020-07-30
Encoding wildcard permissions automatically and explicitly


niomaster:
I had an idea on how we might automatically encode wildcard permissions. Often we encounter limitations of wildcard permissions, where either we can't prove the wildcard permissions somehow, or we really want to have the same amount of permission back, but can't. In both cases, we must rewrite code like:

```
context Perm(x.field, read);
void test(x: Object);

void use(Object x) {
    test(x);
}
```

to:

```
given frac p;
context Perm(x.field, p);
void test(x: Object);

void use(Object x) {
    test(x) with { p = 1/2 };
}
```

It would be nice if we can do this transformation automatically. Wildcard permission specifications can appear pretty much anywhere in an expression, which presents a problem in for example:

```
context (\forall int i; l <= i && i < h; Perm(x[i], read));
void sort(int[] x, int l, int h);

void use(int[] x) {
   sort(x, l, h);
}
```

I think we can solve this with maps. We might transform the sort method to something like:

```
given map<int, frac> ps;
requires (\forall int i; l <= i && i < h; (i in ps && ps[i] > 0));
context (\forall int i; l <= i && i < h; Perm(x[i], ps[i]));
void sort(int[] x, int l, int h);
```

At the usage site we can use the "current permission value" feature in silver to ensure we don't give away all permission:

```
void use(int[] x) {
    map<int, frac> ps;
    assume (\forall int i; l <= i && i < h; (i in ps && ps[i] == perm(x[i]) \ 2));
    sort(x, l, h) with {ps=ps};
}
```

This assumption is sound, because it only deals with permission values. The requirements of sort then prove the values are >0.

Note that x, l and h in the assume statement refer to the argument values to sort, they're just the same here to illustrate.

bobismijnnaam:
This sounds very useful. We can only do this for permissions of the form `Perm(_, read)`, right? Otherwise we'd have to do a tricky encoding I think (see [here](https://ethz.ch/content/dam/ethz/special-interest/infk/chair-program-method/pm/documents/Education/Theses/Benjamin_Schmid_BA_report.pdf)).

---------------------------

One thing (as discussed earlier with Pieter, just documenting it here) is the following problem. There might be an "unsound typo" in a requires:

```java
context (\forall int i; l <= i && i < h; Perm(x[i], read) ** false);
void sort(int[] x, int l, int h);
```

If the transformation is very naive and I understand it correctly, calling `sort` will produce the following `assume`:

```java
[...]

PR #1207, opened 2024-05-24
Add AutoValue

superaxander:
Checklist:

# PR description
Allows you to have a method that needs read permission to some field but doesn't modify the callers permission amount. (Unlike `Perm(loc, read)`/`Value(loc)` which result in the loss of write permission for the caller since the callee need not prove that it returns the same amount of permission it took)

This is related to #502 but it is not replacement for the idea described there since this implementation is more restricted. 

If we have a method:
```java
context AutoValue(a.b);
void doStuff(A a) {
    // Read from a.b
}
```

This is transformed into
```java
requires \polarity_dependent(Perm(a.b, read$()), !(\forperm A x \in x.b; false) ** !(\forperm A x \in x.b; x != a));
ensures \polarity_dependent(!(\forperm A x \in x.b; false) ** !(\forperm A x \in x.b; x != a), Perm(a.b, read$()));
void doStuff(A a) {
    // Read from a.b
}
```

Which essentially checks at the call site if the caller has access to the field and then inside the method we inhale some symbolic amount of permission (greater than 0 but less than 1) which has to be exhaled at the end of the method again.

Inhaling the `forperm` stuff in the postcondition is only necessary to be able to write other postconditions which require access to the field, it doesn't make any difference for the caller's heap. 

Unless the expression containing `AutoValue` is in the both the precondition and the postcondition (like is the case when you use `context`) an automatic leak check is added such that you cannot leak some part of the symbolic read permission which would make this unsound. In the above example (assuming instead of `context` we wrote `requires`) this leak check would look like this:

```java
ensures \polarity_dependent(true,Perm(a.b, read$()));
```

If the leak check fails there is an appropriate error returned to the user. (`autoValueLeak:perm`) An example of such a leak is given in https://github.com/superaxander/vercors/blob/b3396287f2fc91667df094138ccbdf90664e2d79/examples/concepts/autovalue/leak.pvl

Currently you can use `AutoValue` only in expressions containing `**`, `let`, `? :`, and `==>`. For `let`, `? :` and `==>` an implicit `\old` is added around the parts of the expression which do not contain the AutoValue when used in the postcondition such that you cannot leak the permission by for example having an expression `boolean ==> AutoValue(b.a)` where boolean is true when the method is called but false when it returns. 

An alternative way to encode the assertion that there exists read permission (instead of the `forperm` expressions) for the given field could be `perm(loc) > 0`. Intuitively this sounds like it might be more performant but I have not done any testing, nor do I know enough about how Viper works to evaluate that.

### DIFF
--- a/src/rewrite/vct/rewrite/EncodeAutoValue.scala
+++ b/src/rewrite/vct/rewrite/EncodeAutoValue.scala
@@ -1,0 +1,221 @@
+package vct.rewrite
+
+import hre.util.ScopedStack
+import vct.col.ast._
+import vct.col.compare.Compare
+import vct.col.origin.{AbstractApplicable, AutoValueLeakCheckFailed, Blame, LabelContext, Origin, PanicBlame, PostBlameSplit, PostconditionFailed, PreferredName, TrueSatisfiable}
+import vct.col.util.AstBuildHelpers._
+import vct.col.rewrite.{Generation, Rewriter, RewriterBuilder, Rewritten}
+import vct.result.VerificationError.UserError
+
+import scala.collection.mutable
+
+case object EncodeAutoValue extends RewriterBuilder {
+  override def key: String = "encodeAutoValue"
+
+  override def desc: String = "Encodes the AutoValue permission value"
+
+  private def AnyReadOrigin: Origin = Origin(
+    Seq(
+      PreferredName(Seq("read$")),
+      LabelContext("encodeAutoValue")
+    )
+  )
+
+  case class InvalidAutoValue(e: Expr[_]) extends UserError {
+    override def code: String = "invalidAutoValue"
+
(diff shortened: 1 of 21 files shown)


superaxander:
All the tests seem to pass except for the one that is already failing on dev

src/rewrite/vct/rewrite/EncodeAutoValue.scala
```diff
@@ -0,0 +1,219 @@
...
+            Let(b, v, m)
+          } else {
+            mMap.foreach(postM => conditionContext.top._2.append((Let(binding, value, postM._1), Let(b, v, postM._2))))
+            conditionContext.top._1 match {
+              case InPrecondition() => Let(b, v, m)
+              case InPostcondition() => Let(b, Old(v, None)(PanicBlame("Old should always be valid in a postcondition")), m)
+            }
+          }
+        }
+        // This one is probably useless since you can't use resource expressions with and...
```
superaxander: Oh I forgot to get rid of this comment

PR #1212, opened 2024-06-05
Fix build on JDK 21+, resolves #1146

### LINKED ISSUE #1146: Building VerCors on Java 21 results in package ops not found error
When trying to build Java usin Java 21, mill/scalac reports an error that for `col`, the `ops` package cannot be found. Subsequently removing the `out/` folder and then building VerCors with Java 17 makes the problem go away. This happened on commit bfd486d842d92d08bdde1d2156b9faca66115872.

(@superaxander please let me know if I got any details wrong.)

superaxander:
Files were not being generated leading to the missing packages in #1146 

It seems on newer JDKs there is no $ after Lambda so that caused our name matching to fail.

Example of what a class name looks like now:
```
scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances$$Lambda/0x00007f00eac376f8
```

### DIFF
--- a/src/helpers/scala/meta/ResultStream.scala
+++ b/src/helpers/scala/meta/ResultStream.scala
@@ -108,7 +108,7 @@ object ResultStream {
       case Function(fn) =>
         if (
           fn.getClass.getName.startsWith(
-            "scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances$$Lambda$"
+            "scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances$$Lambda"
           )
         ) {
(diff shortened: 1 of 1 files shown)


superaxander:
I was wondering why the ??? didn't cause an exception which would've failed the generation phase of the build, but it happens inside a Using where the resulting Try is ignored

pieter-bos:
Yeah I've just misunderstood how `Using` works, likely all uses of `Using` should be replaced with `Using.resource`. Like I mentioned recently these would have been catched by `-Wvalue-discard`, but it has a good number of false positives for our code.

PR #1238, opened 2024-09-16
Remove unused nlohmann/json dep for vcllvm.origin module

wandernauta:
Previous versions of this module used the library, but it appears the current version does not, so this can be removed.

There is no change to behavior.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -673,30 +673,8 @@ object vercors extends Module {
       )
     }
 
-    object json extends LinkableModule {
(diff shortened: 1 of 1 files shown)


superaxander:
This is a change that I've already made in #1159 along with a lot of other changes. I would prefer making this change only there so I don't have to deal with another merge conflict on build.sc

wandernauta:
Ah, right. I'll close this one.

PR #1239, opened 2024-09-16
Encode that malloc may return NULL in C (weaken postcondition)

### LINKED ISSUE #1233: VerCors optimistically assumes malloc always succeeds in C
The following C program verifies:

```c
#include <assert.h>
#include <stdlib.h>

int main() {
    int* ptr = (int*) malloc(sizeof(int)*1000000);
    assert(ptr != NULL);
    free(ptr);
}
```

However, `malloc` can fail, see e.g. section 7.22.3.4 in [ref](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf):

> The malloc function returns either a null pointer or a pointer to the allocated space.

So the assertion doesn't necessarily hold (here on Linux, glibc):

```
$ ( ulimit -v 5000 ; ./malloc_optimism )
malloc_optimism: malloc_optimism.c:6: main: Assertion `ptr != NULL' failed.
Aborted (core dumped)
```

In #973, the `@ensures` clause was given as only promising a pointer to some memory if the result is non-NULL, but it appears that `LangCToCol` and the `concepts/c/malloc_free.c` example currently assume that allocation always succeeds.

---

This may be a known limitation, but it was surprising to me so I thought I'd report it here.

wandernauta:
The new field 'fallible' on the NewPointerArray AST node stores whether you are guaranteed to get a new pointer array (false) or merely probably get one (true).

Updates tests, fixes #1233.

### DIFF
--- a/examples/concepts/c/malloc_free.c
+++ b/examples/concepts/c/malloc_free.c
@@ -13,6 +13,7 @@ struct e{
 
 int main(){
     int* xs = (int*) malloc(sizeof(int)*3);
+    if (xs == NULL) return 1;
 
     xs[0] = 3;
     xs[1] = 2;
@@ -20,6 +21,7 @@ int main(){
     free(xs);
 
     int** xxs = (int * *) malloc(sizeof(int *)*3);
+    if (xxs == NULL) return 1;
 
     int temp[3] = {1,2,3};
     xxs[0] = temp;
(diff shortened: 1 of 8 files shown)


sakehl (dismissed):
Seems reasonable to me. When implementing malloc I indeed did not consider to have it possibly fail.

superaxander (dismissed):
This looks good, I'll probably replace this with the version from my branches later though where I have a separate `NonNullPointer` type which I need for use in some axioms. (it becomes a `Pointer` type in Viper instead of an `Option[Pointer]`)

wandernauta:
Rebased to resolve conflict with my own changes to `free`.

PR #1251, opened 2024-09-27
Bump dorny

bobismijnnaam:
# PR description

Bump dorny/test-report in the hope it will start working again.


superaxander:
Haha I was just looking into this. I think we actually need to do what they suggest in their readme where the test reporter is a separate workflow not running on the PR branch

superaxander:
Note that the test reports aren't broken on your PRs since they are branches in this repository and hence have access to the resources

bobismijnnaam:
Ah yeah, that sounds logical! I'll delete this branch & PR then.

PR #1260, opened 2024-10-14
Add stringToName special case for all-underscores identifier strings

wandernauta:
Before, parsing the following PVL program like the following would give an assertion error, since it would attempt to build a preferred name with no parts:

```php
void ___() {
}
```

It now gets a preferred name with one part, the string "___".

Also adds a unit test for stringToName.

### DIFF
--- a/src/col/vct/col/origin/Origin.scala
+++ b/src/col/vct/col/origin/Origin.scala
@@ -89,7 +89,9 @@ case class IndirectName(name: Name) extends NameStrategy {
 object SourceName {
   def stringToName(name: String): Name =
     Name.Preferred(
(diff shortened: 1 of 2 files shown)


superaxander:
Looks like the compilation fails because github updated the `ubuntu-latest` runner image from `ubuntu-22.04` to `ubuntu-24.04` which contains a newer version of LLVM while VCLLM still expects LLVM 15 to be present.  Once #1159 is merged (hopefully this week) this will be resolved.

wandernauta:
Indeed. I'll rebase this (and #1261) on `dev` once that PR is in. It looks really cool!

PR #1261, opened 2024-10-14
Collect bip_annotation modifier in all ToCol classes

wandernauta:
The bip_annotation modifier is part of the specification grammar for all input languages, so a PVL program like this parses according to ANTLR:

```
bip_annotation
void foo() {
}
```

...but then hits a MatchError ("VerCors has crashed") in the PVLToCol transform.

This PR changes PVLToCol, CToCol and CPPToCol to follow the same strategy as LLVMContractToCol, which is to pass through the BIP annotation. (Actually, the methods are now identical.) In practice, this results in a "modifier cannot be attached to this declaration" message in the example above.


bobismijnnaam:
Nice catch, thanks :)

PR #1278, opened 2024-11-13
Log and exit error if --compile (PVL-to-Java translation) mode fails

wandernauta:
Before, the mode would silently return success (0) even if compilation actually failed, for example because of a syntax error. This brings behavior more in line with the other modes.

### DIFF
--- a/src/main/vct/main/modes/Compile.scala
+++ b/src/main/vct/main/modes/Compile.scala
@@ -8,6 +8,7 @@ import vct.main.Main
 import vct.main.stages.{Output, Parsing, Resolution, Transformation}
 import vct.options.Options
 import vct.parsers.transform.ConstantBlameProvider
+import vct.result.VerificationError
 
 import java.nio.file.Paths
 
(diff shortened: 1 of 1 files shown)


bobismijnnaam:
Looks good. If you don't mind changing the PR to also print the systemerror when it is encountered, and return EXIT_CODE_ERROR in that case, the PR can be merged.

By the way, general request: could you leave a comment on the PR when you're finished? Most of your PRs are so small I can never quite tell if you're finished or not, so it would be helpful if you could mark your  PR as officially finished after the last PR. Or you could mark it as draft when you create the PR, and then unmark it as draft when you're finished. Whatever you prefer. Other than that, keep up the good work :)

wandernauta:
Thanks for taking a look! I do try to keep PRs as small as possible :) In future, I'll create PRs as draft and un-draft them when I think they're ready.

In practice, system errors are already printed and exit-1'ed because the exception bubbles to `vct/main/Main.scala:53`, which does the whole "VerCors has crashed" banner and GitHub link - the same as Verify mode, basically.

Would you like to instead see that system errors are caught inside Compile.scala itself, so that that banner is not printed, only the error itself?

bobismijnnaam:
Ah, you're right. LGTM

PR #1309, opened 2025-02-11
Quantifier fix

sakehl:
Checklist:

Should not be needed

# PR description

After trying to formally verify the quantifier rewrite pass in Lean4, I realized there were some errors there. Mostly with respect to negative numbers.

For instance the quantifier:
`(\forall int x=xmin..xmin+nx, int y=ymin..ymin+ny, int z=zmin..zmin+nz;  xs[a0*x+a1*y+a2*z+b]>0)` 

should be rewritten towards
`off = b + a0*xmin + a1*ymin + a2*zmin`
```
(\forall int xyz;  
  (a0>0 ==> 0 <= xyz-off && xyz-off<a2*nz) &&
  (a0<0 ==> a2*nz < xyz-off && xyz-off<= 0)&&
  |xyz - off| % a2 % a1 % a0 == 0;
  xs[xyz]>0
```
But we also need to determine that `a0!=0`, `nx>0`, `ny>0`, `nz>0`, `a1==nx*a0`, `a2==ny*a1`. 

Other conditions could be valid as well, but then we need to determine if `sign(a0)=sign(a1)==sign(a2)`. So I've added a few ways to symbolically determine if signs are equal or not.

These changes should let the quantifier pass have a formally correct basis. I've also added some tests to determine if quantifiers are correctly rewritten.


sakehl:
@superaxander or @bobismijnnaam 
Would one you be willing to have a quick look at it, to see if I can merge this.

The changes are all local to the simplify nested quantifiers pass / expression equality check class

superaxander:
That stack overflow in the tests with isConstantIntRecurse seems like a problem with this PR

sakehl:
Lol, it crashed my "Should Not crash with stack overflow" test. Anyway, should be fixed now:)

superaxander:
Looks good to me, the remaining failed tests are unrelated to this PR

bobismijnnaam:
Cool Lars! LGTM

PR #1311, opened 2025-02-12
Rewriting the SYCL encoding

OmerSakar:
Checklist: 

# PR description


bin/silicon
```diff
@@ -3,10 +3,10 @@ set -e
 ROOT="$(dirname "$(dirname "$(readlink -f $0)")")"
 (cd "$ROOT"; ./mill -j 0 vercors.main.runScript)
 
-if [[ "$OSTYPE" == "darwin" ]]; then
+#if [[ "$OSTYPE" == "darwin" ]]; then
```
superaxander: You probably didn't mean to commit this :)

res/universal/res/c/emmintrin.h
superaxander, line 1: Did you change anything in this file? It seems like it was auto-formatted but it does not look great

res/universal/res/c/opencl.h
superaxander, line 1: Same for this file

res/universal/res/cpp/sycl/sycl.hpp
```diff
@@ -1,5 +1,28 @@
 namespace sycl {
 
+  namespace h {
```
superaxander, line 3: Will users ever need to interact with this namespace, if so why is it called `h`?

res/universal/res/cpp/sycl/sycl.hpp
```diff
@@ -1,5 +1,28 @@
...
+    /*@
+      ensures (a>0 && b>0) ==> \result > 0;
+      ensures \result == a*b;
```
superaxander: This one should be unnecessary, unless you want to make this function opaque. (Same goes for the `add` function)

src/col/vct/col/typerules/CoercingRewriter.scala
```diff
@@ -1269,13 +1270,16 @@ abstract class CoercingRewriter[Pre <: Generation]()
...
-        CPPClassMethodOrFieldAccess(classInstance, methodOrFieldName)(cfa.blame)
+        CPPClassMethodOrFieldAccess(classInstance, methodOrFieldName, typeArgs)(
+          cfa.blame
+        )
       case defn @ CPPLambdaDefinition(contract, declarator, body) =>
         CPPLambdaDefinition(contract, declarator, body)(defn.blame)
       case CPPLambdaRef() => e
-      case inv @ CPPInvocation(applicable, args, givenArgs, yields) =>
-        CPPInvocation(applicable, args, givenArgs, yields)(inv.blame)
+      case inv @ CPPInvocation(applicable, args, givenArgs, yields, sginv) =>
```
superaxander: You might want to coerce the sub-group invariant to resource in case someone enters a non-resource expression in there, or is that checked somewhere else already?

src/llvm/lib/Util/PallasWrapperUtils.cpp
superaxander, line 1: When you merge this with `dev` take care not to change anything when resolving conflicts

src/main/vct/main/stages/Resolution.scala
```diff
@@ -3,19 +3,10 @@ package vct.main.stages
...
-  VCLLVMFunctionContract,
-  LLVMGlobalSpecification,
-  Program,
-  Verification,
-  VerificationContext,
-}
+import vct.col.ast.{ApplicableContract, Expr, Function, GlobalDeclaration, LLVMGlobalSpecification, Program, VCLLVMFunctionContract, Verification, VerificationContext}
 import vct.col.check.CheckError
 import vct.col.origin.{FileSpanningOrigin, Origin, ReadableOrigin}
+import vct.col.print.Ctx
```
superaxander: Unused import?

src/main/vct/main/stages/Resolution.scala
```diff
@@ -32,14 +23,9 @@ import vct.parsers.ParseResult
...
-import java.io.{
-  InputStreamReader,
-  OutputStreamWriter,
-  StringReader,
-  StringWriter,
-}
+import java.io.{InputStreamReader, OutputStreamWriter, StringReader, StringWriter}
 import java.nio.charset.StandardCharsets
-import java.nio.file.Path
+import java.nio.file.{Path, Paths}
```
superaxander: Unused import?

src/main/vct/main/stages/Resolution.scala
```diff
@@ -32,14 +23,9 @@ import vct.parsers.ParseResult
 import vct.resources.Resources
 import vct.result.VerificationError.UserError
 
-import java.io.{
-  InputStreamReader,
-  OutputStreamWriter,
-  StringReader,
-  StringWriter,
-}
+import java.io.{InputStreamReader, OutputStreamWriter, StringReader, StringWriter}
```
superaxander, line 25: Unused import?

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2,31 +2,27 @@ package vct.rewrite.lang
...
-  _,
-}
 import vct.col.origin._
 import vct.col.ref.Ref
 import vct.col.resolve.NotApplicable
 import vct.col.resolve.ctx._
 import vct.col.resolve.lang.CPP
 import vct.col.rewrite.ParBlockEncoder.ParBlockNotInjective
-import vct.col.rewrite.{Generation, ParBlockEncoder, Rewritten}
+import vct.col.rewrite.{Generation, ParBlockEncoder, Rewriter, Rewritten, SimplifyNestedQuantifiers}
```
superaxander: Unused imports?

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2,31 +2,27 @@ package vct.rewrite.lang
...
 import vct.col.util.AstBuildHelpers.{assignLocal, _}
-import vct.col.util.{AstBuildHelpers, SuccessionMap}
+import vct.col.util.{AstBuildHelpers, Substitute, SuccessionMap}
```
superaxander: Unused import?

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2,31 +2,27 @@ package vct.rewrite.lang
...
 import vct.result.Message
 import vct.result.VerificationError.{Unreachable, UserError}
 import vct.rewrite.lang.LangSpecificToCol.NotAValue
 
+import scala.Seq
```
superaxander: Unused import?

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2,31 +2,27 @@ package vct.rewrite.lang
...
 import scala.collection.mutable
+import scala.runtime.BoxedUnit
```
superaxander: Unused import

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -210,6 +214,14 @@ case object LangCPPToCol {
       inv.blame.blame(SYCLItemMethodPreconditionFailed(error.node))
   }
 
+  private case class SYCLMulMethodInvocationBlame()
```
superaxander: Why introduce this blame when you can simply pass in the PanicBlame directly?

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -706,72 +790,68 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
   val syclLocalAccessorVarToDimensions
       : mutable.Map[Variable[Post], Seq[Expr[Post]]] = mutable.Map.empty
 
-  val currentlyRunningKernels
-      : mutable.Map[Local[Post], Seq[SYCLAccessor[Post]]] = mutable.Map.empty
+  var syclBufferRangeRelations: Seq[Eq[Post]] = Nil
```
superaxander, line 803: Haven't looked further down yet, but this might make more sense as a ScopedStack? Or can there only ever be a single method with SYCL kernels? It seems weird to me that this is a global thing

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -706,72 +790,68 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+
+  val currentlyRunningKernels: mutable.Map[Local[
```
superaxander, line 805: Similarly for this one it sounds to me like this doesn't make sense outside some scope

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
superaxander: If you need to use this in multiple places you could just define it as `def SYCLMulMethodInvocationBlame = PanicBlame(...)`

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1409,20 +1534,136 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+      case SYCLTNDRange(_) =>{
+        syclWarpSize = Some(new Variable[Post](TCInt())(kernelDeclaration.o.where(name="warpSize")))
+        //        rw.variables.declare(syclWarpSize.get)
+
+        val maybeWarpSize = kernelDeclaration.declarator.asInstanceOf[CPPLambdaDeclarator[Post]].attrs
+          .find(attr => attr.name == "sycl::reqd_sub_group_size" && attr.args.size == 1 && attr.args.head.t == TCInt[Post]())
+
+        val defineWarpSize = maybeWarpSize match {
+          case Some(ws) => (
+            Assign(Local[Post](syclWarpSize.get.ref)(ws.o), ws.args.head)(PanicBlame("The assignment of the warpsize should never fail"))(ws.o.where(name="warpSize")),
```
superaxander: We have a predefined PanicBlame for assigning locals `AssignLocalOk`, there's also the helper function `assignLocal(local, value)`

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1409,20 +1534,136 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+            Eq(Local[Post](syclWarpSize.get.ref)(ws.o), ws.args.head)(ws.o),
+          )
+          case None =>
+            implicit val o: Origin = invocation.o
+            val wrpsizVar = new Variable[Post](TCInt())(o.where(name="warpSize"))
+            val wrpsizeProcedure = withResult((result: Result[Post]) => {
+                new Procedure[Post](
```
superaxander: Since your using default arguments here a lot I recommend using the `procedure` helper function from ASTBuildHelpers

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1409,20 +1534,136 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+                    decreases = None,
+                  )(PanicBlame("There are no preconditions"))(invocation.o),
+                  inline = false,
+                  pure = true,
+                  opaque = false,
+                )(PanicBlame("Should never fail"))(o.where(name = "warpSizeFunc"))
+              })
+
+              rw.globalDeclarations.declare(wrpsizeProcedure)
+              val callToWrpsizeProcedure = new ProcedureInvocation[Post](
```
superaxander: Similarly you can use the `procedureInvocation` helper function here

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1409,20 +1534,136 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+                                    )
+      => local.ref.get.asInstanceOf[RefCPPLocalDeclaration[Pre]]
+    }
+
+    val readOnlyPerms = accsInBody.map { accRef =>
+      val buff = syclAccessorSuccessor(accRef).buffer.generatedVar
+      implicit val o = accRef.decls.o
+      val indexVar = new Variable[Post](TCInt())(o.where(name = "i"))
+      val blame = PanicBlame("Should never fail. Please report if it does fail")
+      Starall(
```
superaxander: Here I can recommend using the `starall` helper function which makes the variable for you

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1409,20 +1534,136 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+              Seq(
+                ArraySubscript(
+                  Local[Post](buff.ref),
+                  Local[Post](indexVar.ref)
+                )(blame)
+              )
+            ),
+            body = (Local[Post](indexVar.ref) >= c_const(0)  &&
+                Local[Post](indexVar.ref) < Length(Local[Post](buff.ref))(PanicBlame("Should never be null")))
+            ==> Perm(
```
superaxander: The canonical way to write read permission is with the `Value` node

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1409,20 +1534,136 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+      val buff = syclAccessorSuccessor(accRef).buffer.generatedVar
+      implicit val o = accRef.decls.o
+      val indexVar = new Variable[Post](TCInt())(o.where(name = "i"))
+      val blame = PanicBlame("Should never fail. Please report if it does fail")
+      Starall(
+            bindings = Seq(indexVar),
+            triggers = Seq(
+              Seq(
+                ArraySubscript(
+                  Local[Post](buff.ref),
```
superaxander: This pattern seems to be everywhere here, you can do `v.get` instead this is an extension function from ASTBuildHelpers that expands to `Local(new DirectRef(v))`

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1548,15 +1802,15 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
     currentDimensionIterVars.clear()
     syclAccessorSuccessor.clear()
     currentThis = None
-
+    syclBufferRangeRelations = Nil
```
superaxander, line 1792: These are clear candidates for using a ScopedStack

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1592,40 +1839,44 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
       Local[Post](iterVar.variable.ref) >= c_const(0) &&
       Local[Post](iterVar.variable.ref) < iterVar.to
     })
     implicit val o: Origin = kernelDeclaration.o
 
+    gatherBlockStatements = true
+    visitedKernelStatements = mutable.Seq.empty
+    val body = rw.dispatch(kernelDeclaration.body)
+    visitedKernelStatements = mutable.Seq.empty
+    gatherBlockStatements = false
```
superaxander, line 1843: Another good candidate for using ScopedStack

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1860,26 +2105,24 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
-                    val rangeIndexFields = Seq
-                      .range(0, buffer.range.dimensions.size).map(i =>
-                        new InstanceField[Post](TCInt(), Nil)(
-                          dimO.where(name = s"${accName}_r$i")
-                        )
-                      )
+                    val localBuff = buffer.generatedVar.get(accO)
+
+                    val rangeIndexLocal = buffer.range.dimensions.map { d =>
+                      d.asInstanceOf[Local[Post]]
```
superaxander, line 2099: I guess that the condition that buffer.range.dimensions contains only Locals is preserved elsewhere, I am wondering if it'd be possible to just have the field in SYCLAccessor contain Expr's do you need it to be a local? In any case you could add an assertion to the body of SYCLRange to check that it only contains Local's when it is constructed (then you'd get an error where the mistake is made instead of here)

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -1926,159 +2169,41 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
-          )(acc.instanceField.o),
-          ValidArray(
-            Deref[Post](classObj, acc.instanceField.ref)(
-              new SYCLAccessorDerefBlame(acc.instanceField)
-            )(acc.instanceField.o),
-            rangeIndexDerefs.reduce((e1, e2) => (e1 * e2)(acc.buffer.o)),
-          )(acc.instanceField.o),
-        )
-      )(acc.instanceField.o),
+      foldStar(Seq(
```
superaxander: foldStar with only one expression in the sequence

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2422,13 +2623,14 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
 
     // Wait for SYCL kernels that access the buffer to finish executing
     val kernelsToTerminate = currentlyRunningKernels.filter(tuple =>
```
superaxander, line 2625: Could make this a `.filter { case (key, (pred, accessors)) =>` to prevent the _.2._1 stuff down below

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2609,41 +2796,98 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+      e: Expr[Post],
+      accs: Seq[SYCLAccessor[Post]],
   ): Expr[Post] = {
     implicit val o = e.o
     e match {
       case s @ Starall(bindings, triggers, body) =>
-        Starall(bindings, triggers, removeKernelClassInstancePermissions(body))(
-          s.blame
-        )
+        Starall(
```
superaxander, line 2806: could be `s.rewrite(body=removeKernelClassInstancePermissions(body, accs))`

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2609,41 +2796,98 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
         )
       case Wand(left, right) =>
         Wand(
-          removeKernelClassInstancePermissions(left),
-          removeKernelClassInstancePermissions(right),
+          removeKernelAccessorPermissions(left, accs),
+          removeKernelAccessorPermissions(right, accs),
+        )
+
+      case ActionPerm(Local(obj), _)
```
superaxander, line 2838: Could be `case ActionPerm(Local(Ref(v)), _)` replacing `obj.decl` with `v`

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2609,41 +2796,98 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+        )
+      case e => e
+    }
+  }
+
+  private def removeRangeAccesses(e: Expr[Post]): Expr[Post] = {
+    implicit val o = e.o
+    e match {
+      case s @ Starall(bindings, triggers, body) =>
+        Starall(bindings, triggers, removeRangeAccesses(body))(s.blame)
```
superaxander, line 2873: again could be `s.rewrite(body=removeRangeAccesses(body))`

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2794,4 +3030,215 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+          GlobalThreadId[Pre]() -> Plus(getGlobalWorkItemLinearId(inv), d),
+          SubGroupFuncValue[Pre]() -> Local(sglResult.ref),
+        )
+      ) {
+        rw.dispatch(sgInv)
+      }
+    }
+    val warpsize = syclWarpSize.getOrElse(throw new SYCLWarpSizeNotDefinedError(inv))
+
+    val predToExhale = Implies(
```
superaxander: could be `val predToExhale = (c_const(0) < laneid - d) ==> exhalePred`

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2794,4 +3030,215 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+
+    if (dependsIndirectlyOnSYCLIdFunctions(sgInv)) { throw SYCLSubGroupInvDependsOnTheId(sgInv) }
+
+    val sg = new Variable[Post](TSeq(TCInt()))(o.where(name = "sg"))
+    val value = new Variable[Post](TCInt())(o.where(name = "value"))
+    val delta = new Variable[Post](TCInt())(o.where(name = "delta"))
+    val shiftGroupLeftBlame = PanicBlame("The call to shift_group_left should never fail")
+
+    val procedure = withResult((result: Result[Post]) => {
+      new Procedure[Post](
```
superaxander: use `procedure` helper function

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2794,4 +3030,215 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+    rw.globalDeclarations.declare(procedure)
+
+    val result: Expr[Post] = With[Post](
+      Exhale(predToExhale)(SubGroupFunctionPreconditionFailed(inv)),
+      ScopedExpr(
+        Seq(sglResult),
+        Then(
+          PreAssignExpression(
+            Local[Post](sglResult.ref),
+            ProcedureInvocation[Post](
```
superaxander: Use `procedureInvocation` helper function

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
```diff
@@ -2794,4 +3030,215 @@ case class LangCPPToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+          }
+        }.toSet
+        (target, valueSet)
+      }.toSeq
+    // A map of String names of variables -> a set of nodes representing the variables it refers to
+    val varToNames: Map[Object, Set[CPPExpr[Pre]]] = collectNames.map { // Convert the different nodes into their String names for easier lookup
+      case (name: CPPName[Pre], vals) => (name.name, vals)
+      case (local: CPPLocal[Pre], vals) => (local.name, vals)
+      case (a, b) => (a, b)
+    }.groupBy(_._1).map(t => (t._1, t._2.flatMap(_._2).toSet))
```
superaxander: Why add the unnamed things into this map here? It seems like you're just using this to find names anyway?

src/rewrite/vct/rewrite/lang/LangSpecificToCol.scala
```diff
@@ -417,7 +425,12 @@ case class LangSpecificToCol[Pre <: Generation](
       case kernel: GpgpuCudaKernelInvocation[Pre] =>
         c.cudaKernelInvocation(kernel)
       case local: LocalThreadId[Pre] => c.cudaLocalThreadId(local)
-      case global: GlobalThreadId[Pre] => c.cudaGlobalThreadId(global)
+      case global: GlobalThreadId[Pre] if !cpp.syclSubgroupInvSuccessors.isEmpty && cpp.syclSubgroupInvSuccessors.top.contains(global) =>
+        cpp.syclSubgroupInvSuccessors.top(global)
+      case global: GlobalThreadId[Pre] => {
+        c.cudaGlobalThreadId(global)}
+      case sgfv: SubGroupFuncValue[Pre] if !cpp.syclSubgroupInvSuccessors.isEmpty && cpp.syclSubgroupInvSuccessors.top.contains(sgfv) =>
+        cpp.syclSubgroupInvSuccessors.top(sgfv)
```
superaxander: could be `cpp.syclSubgroupInvSuccessors.exists(_.contains(sgfv))` I think

res/universal/res/c/emmintrin.h
OmerSakar: I have reverted it back to match dev.

res/universal/res/c/opencl.h
OmerSakar: I have reverted it back to match dev.

res/universal/res/cpp/sycl/sycl.hpp
OmerSakar: Where should this be documented?

superaxander: Well wherever the documentation for using SYCL in VerCors is :P

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
OmerSakar: s is of type Expr[Post], the rewriter (that rewrite uses implicitly) is Pre -> Post, so this does not work.

res/universal/res/cpp/sycl/sycl.hpp
OmerSakar: Its a question whether I will have time to write documentation for this. The users do interact with it, namely so that some quantifiers trigger (e.g. when using linearize2 for example.

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
OmerSakar: s is of type Expr[Post], the rewriter (that rewrite uses implicitly) is Pre -> Post, so this does not work.

src/rewrite/vct/rewrite/lang/LangCPPToCol.scala
OmerSakar: The variables are associated with the buffer, which can have multiple accessors associated with it, so putting them in SYCLAccessor would not be an option I think.

Adding an assertion to the body of SYCLRange is also not an option, since a range does not have to contain only locals, but in the case of SYCLBuffers, their range does only contain locals. 

I added an unreachable here, which seems fitting, this should never happen in production

PR #1318, opened 2025-02-27
Some fixes to const pointers and clean up variable to pointer

sakehl:
# PR description
## Const pointers

@superaxander pointed out to me that several programs with const pointers, and especially if you take the address of a const pointer, were not working.

Things like this
```c
int f(const int a) {
    const int *b = &a;
    return *b;
}
```

and like 
```c
const int a = 5;

//@ ensures \result == 5;
int foo() {
   const int * b = &a;
   return *b;
}
```

should now work.

For now I disallowed the following:

```c
struct vec {
  const int a;
};

//@ context Perm(v, 1\2);
int f(struct vec v) {
  const int *b = &v.a;
  return *b;
}
```
Since it was to much work.

## VariableToPointer pass
It seemed that this pass was not doing anything with fields anymore actually. So I removed all things related with fields. @superaxander can you check if this is done correctly?

### DIFF
--- a/src/rewrite/vct/rewrite/VariableToPointer.scala
+++ b/src/rewrite/vct/rewrite/VariableToPointer.scala
@@ -186,32 +186,6 @@ case class VariableToPointer[Pre <: Generation]() extends Rewriter[Pre] {
               )(PanicBlame("Initialisation should always succeed"))
             } ++ Seq(dispatch(s.body))),
         )
-        // TODO: Can we remove this? It doesn't seem we ever fill fieldMap anyway
-//      case i @ Instantiate(cls, out)
-//          if cls.decl.isInstanceOf[ByValueClass[Pre]] =>
-//        Block(Seq(i.rewriteDefault()) ++ cls.decl.declarations.flatMap {
-//          case f: InstanceField[Pre] =>
-//            if (f.t.asClass.isDefined) {
-//              Seq(
-//                Assign(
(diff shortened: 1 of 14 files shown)


src/rewrite/vct/rewrite/VariableToPointer.scala
```diff
@@ -151,56 +182,36 @@ case class VariableToPointer[Pre <: Generation]() extends Rewriter[Pre] {
...
-                    fieldMap(f).t.asPointer.get.element,
-                    const(1),
-                    None
-                  )(PanicBlame("Size is > 0")),
-                )(PanicBlame("Initialisation should always succeed"))
-              )
-            } else { Seq() }
-          case _ => Seq()
-        })
+        // TODO: Can we remove this? It doesn't seem we ever fill fieldMap anyway
```
superaxander: Yeah this can be removed, it was never actually running anymore since the type of the instance field is a NonNullPointer nowadays. After I merge this with my PR I'll check if nested structs are still being initialised correctly (since that is what this was supposed to do IIRC)

superaxander (dismissed):
Looks good!

src/rewrite/vct/rewrite/VariableToPointer.scala
sakehl: Removed this piece!

sakehl:
Thanks for having a look, I'll merge it after the tests have passed

PR #1319, opened 2025-02-28
Can now take sizeof typedef again

sakehl:

# PR description
The following program crashed before:
```c
#include <stdlib.h>

typedef int test;

int main(){
    test* x = (test *) malloc(sizeof ( test ) );
    int y = (test) 5.0;
}
```
Because it could not parse `test` in the `sizeof`. Now it works.


superaxander:
Looks good to me

PR #1323, opened 2025-03-07
Allow predicate coercion for unique types + consider pure methods + better error messages

sakehl:

# PR description
So I wanted a few things. Foremost, I wanted this to work:
```c
struct dim {
    int min, extent, stride;
    int flags;
};
/*@
inline resource dim_perm(struct dim *dim, rational p, int i) = 
 Perm(&dim[i], 1\2) **
 Perm(dim[i].min, 1\2) **
 Perm(dim[i].stride, 1\2) **
 Perm(dim[i].extent, 1\2) **
 dim[i].extent > 0
 ;
@*/

struct shape {
    int dimensions;
    struct dim *dim;
};

struct buf {
    struct shape shape;
    /*@unique<0>@*/ int *host;
};

/*@
inline resource buffer_pred(struct buf *buf, rational p, int n_dims) = 
 buf != NULL **
 \pointer_length(buf) == 1 **
 Perm(buf, p) **
 Perm(&buf->shape, p) **
 Perm(&buf->shape.dim, p) **
 buf->shape.dim != NULL **
 \pointer_length(buf->shape.dim) == n_dims **
 Perm(&buf->host, p) **
 buf->host != NULL;
@*/

/*@
 context buffer_pred(out, 1\2, 1);
 context dim_perm(out->shape.dim, 1\2, 0);
 context \pointer_length(out->host) == out->shape.dim[0].extent * out->shape.dim[0].stride;
 context out->shape.dim[0].min == 0 && out->shape.dim[0].extent == 100 && out->shape.dim[0].stride == 1;
 context (\forall* int _0; 0 <= _0 && _0 < 100; Perm(&out->host[_0], 1\1));
 ensures \result == out->host[0];
@*/
int f(/*@unique_pointer_field<host, 1>@*/ struct buf *out) {
  return out->host[0];
}
```

Thus I added coercions for predicates. (I think its bad to have to define a separate predicate for each possible unique type).

## 2. I forgot that methods could be pure
And if they are pure, they can be used in contracts of other methods + in functions, so I want to disallowed a nested coercion (maybe possible but hard to reason about).
Also pure methods need to keep their body.

## 3. better errors:
Indicate which type could not be coerced/gave problems + indicate which nested coercion was not possible.

### DIFF
--- a/src/rewrite/vct/rewrite/TypeQualifierCoercion.scala
+++ b/src/rewrite/vct/rewrite/TypeQualifierCoercion.scala
@@ -1,7 +1,13 @@
 package vct.rewrite
 
 import vct.col.ast._
-import vct.col.origin.{AbstractApplicable, AssignLocalOk, Origin, PanicBlame, TrueSatisfiable}
+import vct.col.origin.{
+  AbstractApplicable,
+  AssignLocalOk,
+  Origin,
+  PanicBlame,
+  TrueSatisfiable,
+}
 import vct.col.rewrite.{Generation, Rewriter, RewriterBuilder}
 import vct.col.typerules.CoercingRewriter
 import vct.col.util.AstBuildHelpers._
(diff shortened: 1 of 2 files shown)


sakehl:
PS: it seems like a lot of changes, but a lot is due to me finally having turned on git hooks, woops sorry!

PR #1338, opened 2025-03-26
Vectors fix & xs->x works for array of structs

sakehl:

# PR description
Vectors:
* For vectors, the division operator was not working properly (since float went to rats, and it didn't type check anymore)
* I removed some spurious coercionmap functions to appear after Cint to int coercions
* I've added a bunch of AVX and AVX2 functions in the header file.

xs->x works for array of structs:
* I now allow `struct v xs[1]; xs->x` to work (it complained about not valid vector type before)

### DIFF
--- a/src/rewrite/vct/rewrite/CFloatIntCoercion.scala
+++ b/src/rewrite/vct/rewrite/CFloatIntCoercion.scala
@@ -1,24 +1,7 @@
 package vct.col.rewrite
 
 import vct.col.ast.`type`.typeclass.TFloats
-import vct.col.ast.{
-  BinExpr,
-  CIntegerValue,
-  Cast,
-  CastFloat,
-  CoerceCFloatCInt,
-  CoerceCIntCFloat,
-  CoerceDecreasePrecision,
-  Coercion,
-  Expr,
-  IntegerValue,
-  TCFloat,
-  TCInt,
-  TFloat,
-  TInt,
-  Type,
-  TypeValue,
-}
+import vct.col.ast._
 import vct.col.origin.Origin
 import vct.col.rewrite.{Generation, RewriterBuilder}
 import vct.col.typerules.CoercingRewriter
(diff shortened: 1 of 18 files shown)


src/rewrite/vct/rewrite/CFloatIntCoercion.scala
```diff
@@ -49,7 +55,7 @@ case class CFloatIntCoercion[Pre <: Generation]()
       // to an arbitrary big float.
       case TCFloat(e, m) => TFloats.ieee754_32bit
       case TFloat(e, m) => TFloats.ieee754_32bit
-      case other => super.postCoerce(other)
+      case other => other.rewriteDefault() // super.postCoerce(other)
```
superaxander: Any reason the old thing is still in a comment?

src/rewrite/vct/rewrite/CFloatIntCoercion.scala
```diff
@@ -58,6 +64,6 @@ case class CFloatIntCoercion[Pre <: Generation]()
       case Cast(e, TypeValue(TCInt())) if e.t.isInstanceOf[TCInt[Pre]] =>
         dispatch(e)
       case CIntegerValue(v, _) => IntegerValue(v)(e.o)
-      case other => super.postCoerce(other)
+      case other => other.rewriteDefault() // super.postCoerce(other)
```
superaxander: Idem

src/rewrite/vct/rewrite/adt/ImportVector.scala
```diff
@@ -413,15 +438,24 @@ case class ImportVector[Pre <: Generation](importer: ImportADTImporter)
...
+                  // We do not want to check for non zeroness
+                  isDivOp = false,
                   isCompareOp = false,
-                  (l, r) => FloatDiv[Post](l, r)(div.blame)(o),
+                  // Floats are turned into rationals, and floats don't care about division by zero per se
+                  (l, r) =>
+                    Select(
+                      Neq(r, const[Post](0)),
+                      RatDiv[Post](l, r)(div.blame)(o),
+                      getNonDetFloat(),
```
superaxander: Won't this mean that `1f/0f == 1f/0f` because nonDetFloat is a function? But in reality a NaN always compares false with another NaN (even if they're represented the same) Not that I have a great idea for how to resolve this, but I think we should at least document it

sakehl: Ow yeah that is true, honestly, I did not know that!

Although looking at the specification 1.0/0.0 actually returns infinity (and -1.0/0.0 or 1.0/-0.0 returns negative inf)
See: https://en.wikipedia.org/wiki/IEEE_754#Special_values

So still not completely correct. (And also: I define 'special' values twice now, once in importVector and once in float to Rat. maybe I do change that though)

The proper way to do this, I think, is to actually encode floats/rationals in an ADT, meaning you have:
NormalFloat, signaling/quiet +/- NaN, +/-Inf, +/- 0 and then have operations on floats actually respect these semantics.



Anyway, I'll just add documentation here.

sakehl:
So I was trying to hack NaNs and infs without ADT, and I think that just not works. You really need an ADT, otherwise it to easy to make a mistake in the translation.

Anyway I removed most stoff about NaN again, since NaN propogate through all operations, thus meaning you'd need to check everywhere if things are not NaN, but then you have to make sure all NaNs get handled well etc etc. So I didn't want to open that can of worms completely, so left it at this.

superaxander (dismissed):
Besides the failing test(s) this looks good to me now

sakehl:
Thanks for your review! I had to further dumb down the implementation. E.g. not adding for a/b/ : 'a!=0 && b==inf ==> 0', since when doing that it it actually has no clue that inf is a special value. So even for 5/2 it thinks potentially 2 is inf. so 5/2 does not have a good value anymore.

PR #1347, opened 2025-04-21
Change behavior of nested quantifier rewrite

sakehl:
After some consideration I've changed the behavior of the rewrite nested quantifier pass.
Reason:
* It is bad when 'something magical' happens, where the user has no control over. @bobismijnnaam mentioned this as well before.
 So what I did before is:
  * search for a linear access pattern, when there was no trigger present. And come up with a trigger when that linear access pattern was found in an array index/pointer index/sequence index.
  * For example. `\forall int x1, int x2; 0<= x1 && x1 < 2 && 0 <= x2 && x2<10; Perm(&[xs[x1+x2*2]), write)` was rewritten towards `\forall int x1x2; 0<= x1x2 && x1x2 < 2*10; Perm(&{:[xs[x1x2]:}), write)`.

So what I propose now, is a user has to specifically mark _what_ they want as trigger. And if that trigger contains arithmetic, we try to rewrite this. I reused the trigger pattern, because before this we could never write a trigger with arithmetic before. So it is not the case that old programs were relying on this. And I think this would exactly behave as user was expecting.

This has the benefit that:
* the user tells vercors exactly what he want vercors to rewrite. And if this rewrite fails, vercors fails.

But maybe an even bigger benefit, we can now give more structure to a trigger, then solely an array index. So the following now works:

```java
pure int f(int x, int y);

void main(){
  int n;
  int m;
  assume n>0 && m>0 && n%2==0 &&
  (\forall int x1, int x2, int y;
   0 <= x1 && x1 < 2 && 0 <= x2 && x2 < n/2 && 0 <= y && y < m;
    {:f(x1+2*x2, y):} > 0);
  int a, b;
  assume (a>= 0 && a < n/2) &&
         (b >= 0 && b < m);
  assert f(1+2*a, b) > 0;
  
  int c;
  assume c>=0 && c<n; 
  assert f(c, b) > 0;
   
}
```

I think this is very neat! Since, now if you have arithmethic in a trigger, you can just let vercors safely rewrite this, instead of doing this yourself. But you have control when it does this!

So this is now done. But I've added a few driveby fixes/feautures I also wanted.

* When a GPU kernel says  `requires gridDim.y==1`, we now pick this up, and never parallelize over the y dimension there anymore
  * Reasoning: I think, with quantifiers, I spend a lot of time removing the extra information that the extra parallel loops introduce. Instead of adding a lot of complexity, and trying to remove this extra complexity later on is bad. I think it is better to never introduce this complexity in the first place.
* A special pass for quantifiers with arithmetic and only 1 variable. 
  *Reasoning: When you have a quantifier like `(\forall int i; i % 10 ==0; {:f(5*i+7))` you can always rewrite this by replacing `i` by `(i_new/5-7)` and replacing `5*i+7` by `i_new` and additionally requiring `(i_new-7)%5`. (the only requirment is that the factor in front of the i is not zero). Thus I've added this special case. For my original pass I needed a lowerbound and uperbound for `i`.
* Pass down information from higher contracts.
  *  Reasoning: Before, for the quantifier rewriter to work well, you had to infere a lot of information, which was only available when it was added to the contract. But if a higher contract says `n>0` and you can detect we never write to `n` anyway, we can reuse this information. So we do now.
* Call SMT solver when reasoning
  * Some information, an SMT solver can very easily find, whilst we really struggle. So I've added calls to an SMT solver to determine properties in simplify nested quantifiers. Note I had this laying around forever in an old branch, since I needed it to for some HaliVer examples.   
* I introduced rewrite from a Dafny paper, @bobismijnnaam mentioned before. So if we see. `(\forall int i; ... {:f(i, n+10)):} ... )` we now introduce a new quantifier variable for  `n+10` so we can write a trigger.
* I've added support for opaque functions (see [Release 2024.1](https://github.com/viperproject/silver/blob/master/ReleaseNotes.md). So opaque means that the function body is not visible when calling the function (only the pre and post). Hiding function implementation had fantastic result on an optimised blur example I verified. A quick example is the following:
```java
opaque pure bool f(){
  return true;
}

void test1a(){
  [/expect assertFailed:false] assert f(); [/end]
}

void test1a(){
  assert reveal f();
}
```

That's it!


bobismijnnaam:
Cool!

sakehl:
@superaxander 

Let me know if you want to have a look at this, or someone else, but it is ready for review/merge!

sakehl:
(although I first have to wait on the lastest tests)

examples/concepts/gpgpu/dynamic_shared_cuda.cu
```diff
@@ -29,9 +29,9 @@ __global__ void blur_x(int* in, int* out, int n) {
   }
 
   /*@
-    context Perm(in[0], write \ (blockDim.x * gridDim.x));
-    context blockIdx.x * blockDim.x + threadIdx.x<n ==> Perm(out[blockIdx.x * blockDim.x + threadIdx.x], write);
-    context blockIdx.x * blockDim.x + threadIdx.x<n ==> \old(out[blockIdx.x * blockDim.x + threadIdx.x]) == out[blockIdx.x * blockDim.x + threadIdx.x];
+    context Perm(&n[0], write \ (blockDim.x * gridDim.x));
```
superaxander: Should this be `&in[0]`? (Or rather `in[0]`)

src/col/vct/col/ast/Node.scala
```diff
@@ -735,6 +735,7 @@ final class Function[G](
     val contract: ApplicableContract[G],
     val inline: Boolean = false,
     val threadLocal: Boolean = false,
+    val opaque: Boolean = false,
```
superaxander, line 738: In the future we might want to think about adding some sort of annotation system similar to Viper into VerCors where every expression and every declaration may be annotated instead of having a bunch of booleans (maybe this could be stored in the origin, although that depends on how careful we are not to mess up the origin in places)

src/col/vct/col/ast/util/ExpressionEqualityCheck.scala
```diff
@@ -40,6 +40,13 @@ class ExpressionEqualityCheck[G](info: Option[AnnotationVariableInfo[G]]) {
   var replacerDepth = 0
   val max_depth = 100
 
+  def usefullConditions(): Set[Expr[G]] = {
```
superaxander: Should be `useful` instead of `usefull`

src/rewrite/vct/rewrite/SimplifyNestedQuantifiers.scala
```diff
@@ -1045,60 +1576,211 @@ case class SimplifyNestedQuantifiers[Pre <: Generation]()
...
+          val config = options.build()
+          val logManager = LogManager.createNullLogManager
+          val shutDown = ShutdownNotifier.createDummy
+          var id = 0
+
+          Using(SolverContextFactory.createSolverContext(
+            config,
+            logManager,
+            shutDown,
+            Solvers.SMTINTERPOL,
```
superaxander, line 1618: Any specific reason you chose SMTINTERPOL?

superaxander:
Did a quick review, also sorry for all the changes you'll have to make to make all the perm expressions correct again 🙈

src/rewrite/vct/rewrite/SimplifyNestedQuantifiers.scala
sakehl: I think the other place we use this package we also use SMTINTERPOL, thus I just copied it from there.
But I experimented briefly with others. For SMTINTERPOL and PRINCESS (? I think) you can use this out of the box (it will get the correct packages since these are written in Java I believe. For the other options you have to setup the solvers your self in your build system. So was to much hassle.

src/col/vct/col/ast/Node.scala
sakehl: Hmm I think a good point, but maybe not store in the origin, I think we are often to fussy about origins?

sakehl:
Thanks for the quick review! I accidently removed the inlinePatternLocation, so that was why so much was failing. I think everything looks okay again.

Let me know if its okay to merge like this.

superaxander:
> Thanks for the quick review! I accidently removed the inlinePatternLocation, so that was why so much was failing. I think everything looks okay again.
> 
> Let me know if its okay to merge like this.

I think `examples/concepts/gpgpu/dynamic_shared_opencl.cl` and `examples/concepts/gpgpu/cuda.cu` still have a problem but other than that I think this is ready to merge.

sakehl:
`examples/concepts/gpgpu/dynamic_shared_opencl.cl` looks fine on my end. I indeed missed the cuda one

superaxander:
> `examples/concepts/gpgpu/dynamic_shared_opencl.cl` looks fine on my end. I indeed missed the cuda one

Still seems to be failing the sat check in CI

sakehl:
Owyeah, figured it out now! Added same permission line twice, woops.

Issue #1343, opened 2025-04-08
Crash report: An AST node or a property of the AST was reached, that should have been excluded by an earlier pass.


dzagieboylo:
# Crash Message
```
An AST node or a property of the AST was reached, that should have been excluded by an earlier pass. Language-specific type, expression or declaration was encountered, which should have been already resolved
  at vct.col.rewrite.error.ExtraNode$.<clinit>(ExcludedByPassOrder.scala:26)
  at vct.col.rewrite.ResolveExpressionSideEffects.$anonfun$dispatch$1(ResolveExpressionSideEffects.scala:443)
  at hre.util.ScopedStack.having(ScopedStack.scala:35)
  at vct.col.rewrite.ResolveExpressionSideEffects.dispatch(ResolveExpressionSideEffects.scala:279)
  at vct.col.ast.ops.rewrite.BlockRewrite.$anonfun$rewrite$1(BlockRewrite.scala:6)
  at scala.collection.immutable.List.map(List.scala:250)
  at scala.collection.immutable.List.map(List.scala:79)
  at vct.col.ast.ops.rewrite.BlockRewrite.rewrite(BlockRewrite.scala:6)
  at vct.col.ast.ops.rewrite.BlockRewrite.rewrite$(BlockRewrite.scala:4)
  at vct.col.ast.Block.rewrite(Node.scala:541)
  at vct.col.ast.ops.rewrite.BlockRewrite.rewriteDefault(BlockRewrite.scala:3)
  at vct.col.ast.ops.rewrite.BlockRewrite.rewriteDefault$(BlockRewrite.scala:3)
  ...
  at vct.main.Main$.main(Main.scala:50)
  at vct.main.Main.main(Main.scala)
```

# Version Information
* VerCors version `2.3.0`
* At commit 4bd0a112c13d35dc64e9b357e063afa5af678e6b from branch `HEAD` (changes=false)

# Arguments
`test.cu`

# File Inputs
<details>
<summary>test.cu</summary>

```cu
//:: cases BasicCuda
//:: tool silicon
//:: verdict Pass

#include <cuda.h>

/*@
    context \pointer_index(a, threadIdx.x, write);
@*/
__global__ void example(int a[], int out[], int len) {
    int tid = threadIdx.x;
    atomicAdd(out, a[tid]);
}

```
</details>

# Full Log
<details>

```
13:36:35.050 [INFO] Start: VerCors (at 13:36:35)
[...]

PR #1350, opened 2025-05-01
Cuda support for Atomics

dzagieboylo:
Checklist: 

# PR description

This addresses #1343 by adding support for `kernel_invariant` and `GpgpuAtomic` translation.

This nests the GPU kernel's outer `ParBlock`  inside a `ParInvariant` statement. All `kernel_invariant` expressions are declared as the invariant for this region and are added as *non-scaled* `requires` and `ensures` expressions in the output contract.

The `GpgpuAtomic` is re-written to acquire the declared kernel invariant. This does mean that *all atomic blocks use the same kernel invariant*.

**Note** that this means that thread and block identifiers should not appear in `kernel_invariant` expressions.
I'm unsure what the best way to support that internally would be open to suggestions.

(For example, say you wanted to assume that you have write permissions for `arr[threadIdx.x]` for all thread IDs)

----

Incidentally, I also fixed a bug in which kernel invocations swapped block dimensions and grid dimensions.


src/rewrite/vct/rewrite/SimplifyNestedQuantifiers.scala
```diff
@@ -194,7 +194,7 @@ case class SimplifyNestedQuantifiers[Pre <: Generation]()
   override def dispatch(
       contract: ApplicableContract[Pre]
   ): ApplicableContract[Post] = {
-
+    //TODO check if kernel invariants added correctly
```
sakehl: Looks good

src/rewrite/vct/rewrite/lang/LangCToCol.scala
```diff
@@ -1003,6 +1034,7 @@ case class LangCToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
                 UnitAccountedPredicate(ensures)(o),
                 // Context everywhere is already passed down in the body
                 tt,
+                tt, //TODO determine if we need to do something with kernelinvariants
```
sakehl: Looks reasonable to not have the kernel invariant here anymore.

sakehl (commented):
Looks good as a first implementation.

I think it is also reasonable that thread ids are not allowed to be mentioned in an invariant.

In this older filed is an example how we would want to actually prove something about an atomic: [examples/archive/known-problems/852/GPGPU-Example-updates.cu.](https://github.com/utwente-fmt/vercors/blob/dev/examples/archive/known-problems/852/GPGPU-Example-updates.cu)

Although probably that does not work as is at the moment.

sakehl:
I think probably @superaxander also might want to have a look. And of course the tests should all pass before merging. Unsure what is going on there atm.

Also I think its good if there is a test case testing if the `KernelInvFailure` error is explicitly tested.

And there are going to be a few merge conlflicts with #1347, since I've changed a bit of the GPU encoding (making parallel blocks a bit simpler if block sizes are 1 for the y and z directions)

superaxander:
This looks good to me.

A bunch of the errors in the CI are because the extra kernel invariant field is not filled in in the LLVM frontend (see `/src/llvm/lib/Passes/Function/PallasFunctionContractDeclarerPass.cpp`)  you should be able to copy the `addEmptyRequires`/`addEmptyEnsures` functions to do something similar for the kernel invariant. If you want to compile and test this locally you'll need to have the LLVM 17 libraries installed and add a file called `.include-pallas` to the project root.

dzagieboylo:
Sounds good - I'll make the recommended changes.

I'm happy to wait to merge this until after #1347 and handle the merge conflicts on this PR.

sakehl:
#1347 is merged now, if you need help with merge conflicts shoot me a message, but probably they are not to hard.

dzagieboylo:
After being busy with other things I have come back to handle the LLVM translation, merge, and add some failing tests.
It should be all set now.

dzagieboylo:
> Also I think its good if there is a test case testing if the `KernelInvFailure` error is explicitly tested.

It turns out that this can't ever actually be raised; in the translation I add the invariant as a `requires` condition to the kernel (since there is no actual way to establish the invariant inside the kernel code). This means that the PreCondition checks are actually raised when a kernel invocation is made.

superaxander:
Alright this looks good to me. However, if as you say the KernelInvariantNotEstablished error can never occur then this should probably be a SystemError, i.e. VerCors should crash if it happens and not blame the user. However, I don't know if you, in the future, want to have the ability to do something in the kernel before establishing the kernel invariant. If that is something we want to enable in the future you could leave it as a UserError.

dzagieboylo:
> Alright this looks good to me. However, if as you say the KernelInvariantNotEstablished error can never occur then this should probably be a SystemError, i.e. VerCors should crash if it happens and not blame the user. However, I don't know if you, in the future, want to have the ability to do something in the kernel before establishing the kernel invariant. If that is something we want to enable in the future you could leave it as a UserError.

Done - made it a system error via `PanicBlame`

PR #1351, opened 2025-05-06
OpenCL vector fix, fix for SMT solver query & get info from opencl correctly

sakehl:
Few things:

I realized that many of the opencl vectors had wrong annotations. Indexing was all `...[2*tid+...]` but should be `...[s*tid+...]` for vector of size `s`.

I encoded TruncMod wrongly in the the AsmSMTSolver. I've placed it now in equalityChecker pass, and added a few unit tests to ensure it is correct now.

Lastly opencl was not getting that dimensions were constant, now it does. I've also added to shortcut of using 'get_work_dim()==1`

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -400,7 +400,7 @@ object vercors extends Module {
     }
 
     def key = "col"
-    def deps = T { Agg.empty }
+    def deps = T { Agg(ivy"org.sosy-lab:java-smt:3.14.3") }
     override def sources = T {
       helpers.implTraits.run()
       super.sources()
--- a/examples/concepts/gpgpu/dynamic_shared_opencl.cl
+++ b/examples/concepts/gpgpu/dynamic_shared_opencl.cl
@@ -5,13 +5,13 @@
 #include <opencl.h>
 
 /*@
-  context get_local_size(0) == 32 && get_local_size(1) == 1 && get_local_size(2) == 1;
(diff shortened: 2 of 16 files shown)


sakehl:
I've also placed 'AskSmtSolver' in another class, since that makes more sense, especially if someone want to reuse it somewhere.

I also changed the rewrite rule to no longer split a forall:
```
/* LvdH: I Think it is actually quite bad to split quantifiers here, when a user specified a trigger. Since it removes
   control over the triggering of the quantifier. I think we should not allow this.
axiom forall_and {
  (∀type<any> T, boolean b1, boolean b2;
    (∀T t; (b1!t) && (b2!t)) ==  (∀T t; b1) && (∀T t; b2))
```
I only had to change some triggers from challenge2 from verifythis 2024. The others pass.

I also thought it was bad that someone could write 'get_local_id(4)' whilst we do not support GPU kernels of a bigger dimension size than 3. Thus we throw an error now.

Only the test with 'pointers' does not seem to be always passing anymore. But I guess this was already the case, right?
If I run it on my own device, it passes.

Anyway it is ready to be merged if you agree with the changes here.

superaxander:
Yeah pointer_tag.c is quite unstable unfortunately I've spent quite a long time trying to narrow down why but I'm not super sure yet (although the new viper version should help a bit when we upgrade to that)

These changes look good to me

PR #1355, opened 2025-05-19
Stabilise oddEven permutation example

superaxander:
Checklist: 

# PR description
Applying the trick with more quantified variables to get a good trigger for all the quantifiers. I did this manually but in a pretty straight forward way. I think this could be fully automated. (Probably as an extension of what @sakehl's pass is doing)

There was only one case where the straight forward transformation did not work, there I had to split up the quantifier. (see lemma_odd)

### DIFF
--- a/examples/publications/2020/permutations/oddEven.pvl
+++ b/examples/publications/2020/permutations/oddEven.pvl
@@ -1,24 +1,24 @@
 requires |xs| == |ys|;
 pure boolean property1(seq<int> xs, seq<int> ys) =
-    (\forall int i; 0 <= i && i < |ys|/2 && 2*i+2 < |ys|;
-    ((ys[2*i+1] <= ys[2*i+2]) ==>
-    (xs[2*i+1] == ys[2*i+1] && xs[2*i+2] == ys[2*i+2])));
+    (\forall int i, int j; 0 <= (i-1)/2 && (i-1)/2 < |ys|/2 && i+1 < |ys| && j == i + 1 && i % 2 == 1;
+    ((ys[i] <= ys[j]) ==>
+    ({:xs[i]:} == ys[i] && {:xs[j]:} == ys[j])));
 
 requires |xs| == |ys|;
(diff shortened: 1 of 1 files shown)


examples/publications/2020/permutations/oddEven.pvl
```diff
@@ -1,24 +1,24 @@
 requires |xs| == |ys|;
 pure boolean property1(seq<int> xs, seq<int> ys) =
-    (\forall int i; 0 <= i && i < |ys|/2 && 2*i+2 < |ys|;
-    ((ys[2*i+1] <= ys[2*i+2]) ==>
-    (xs[2*i+1] == ys[2*i+1] && xs[2*i+2] == ys[2*i+2])));
+    (\forall int i, int j; 0 <= (i-1)/2 && (i-1)/2 < |ys|/2 && i+1 < |ys| && j == i + 1 && i % 2 == 1;
+    ((ys[i] <= ys[j]) ==>
+    ({:xs[i]:} == ys[i] && {:xs[j]:} == ys[j])));
```
sakehl, line 5: Pretty sure you can just write 
```
(\forall int i; 0 <= i && i < |ys|/2 && 2*i+2 < |ys|;
    ((ys[2*i+1] <= ys[2*i+2]) ==>
    ({:xs[2*i+1]:} == ys[2*i+1] && xs[2*i+2] == ys[2*i+2])));
```

And it already works and rewrites for you.

superaxander: Yes but I wanted to trigger on both `2*i+1` and `2*i+2` since that greatly reduces the amount of quantifier instantiations (otherwise if you trigger on `2*i+1` that can create a term `2*i+2` which can lead to further triggers, if both of them are triggers it just stops there)

sakehl: Ah nvm, you need to introduce the new quantifier variable j, to trigger on xs[j] as well. That I do not rewrite automatically.

But yes indeed maybe we could do this in the future.

Issue #1301, opened 2025-01-06
Local variable gets initialised in wrong scope


superaxander:
In C we can declare variables inside a loop and take the address of these variables. If we do that, the variable will be replaced by a pointer. In the `VariableToPointer` pass we add a method call to inhale permission to access the pointer at the start of the `Scope` (meaning the `Scope` COL node) where it is declared. Unfortunately, since the `Scope` containing `LocalDecl` is empty until the `CollectLocalDeclarations` pass the `Scope` node is removed by the `PrettifyBlocks` pass. Therefore the permission for the pointer is inhaled in the scope outside of the loop which means that the permission is not available inside the loop since it isn't included in the loop invariant. 

The solution will probably be to either keep the `Scope`s around for longer, automatically re-add the `Scope` if we happen to be inside of a `Loop`, or we move `VariableToPointer` before `CollectLocalDeclarations` and do the initialisation for `LocalDecl`s as well as `Scope`s.

# Crash Message
```
An error condition was reached, which should be statically unreachable. Assigning to a local can never fail.. Inner failure:
 > ======================================
 > At /tmp/vercors-interpreted-17040729019032010151.i
 > --------------------------------------
 >     8  int main() {
 >     9      while (0 == 0) {
 >                   [------
 >    10          int p = 10;
 >                    ------]
 >    11          int *pp = &p;
 >    12          p = 5;
 > --------------------------------------
 > At /home/alexander/vercors-2/test.c
 > --------------------------------------
 >     1  int main() {
 >     2      while (0 == 0) {
 >       [---------------------
 >     3          int p = 10;
 >        ---------------------]
 >     4          int *pp = &p;
 >     5          p = 5;
 > --------------------------------------
 > Insufficient permission to assign to field. (https://utwente.nl/vercors#assignFieldFailed)
 > ======================================
```

# File Inputs
<details>
<summary>/home/alexander/vercors-2/test.c</summary>

```c
int main() {
    while (0 == 0) {
        int p = 10;
        int *pp = &p;
        p = 5;
    }

    return 0;
}

```
</details>

PR #1375, opened 2025-07-28
Fix pointer array which is allocated in loop

sakehl:
Checklist: 

# PR description

The following (contrived) code, does not verify currently.

```c
void main(){
  for(int i = 0; i < 10; i++) {
    float xs[1];
    xs[0] += 1;
  }
  int s=0;
  /*@ loop_invariant 0<=i && i<=10;
   loop_invariant s == i;
  @*/
  for(int i = 0; i < 10; i++) {
    s += 1;
  }
}
```

It seems that the encode pointer array creates an invariant `invariant get_dim_0_float32(xs) == 1` for the second loop invariant, even though xs is not in scope at all.

It seems when we reach the encode pointer array pass, all variables are already declared at the outermost scope, so it does not respect the fact that xs was declared only in the first for loop. Therefore the fix I propose here is: only add the loop invariant when the assignment is still in scope.

Maybe @superaxander can indicate if this is indeed the best idea.

### DIFF
--- a/src/rewrite/vct/rewrite/EncodePointerArrays.scala
+++ b/src/rewrite/vct/rewrite/EncodePointerArrays.scala
@@ -190,10 +190,15 @@ case class EncodePointerArrays[Pre <: Generation]()
     ADTFunction[Post],
   ] = SuccessionMap()
 
-  private val currentVariableContext: mutable.HashSet[Variable[Pre]] = mutable
-    .HashSet()
+  private val currentVariableContext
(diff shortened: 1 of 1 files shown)


superaxander:
I would say the underlying problem is #1301. That variable should never have been in the outer scope but because we get rid of “empty” scopes before we collect all the LocalDecls into the Scope it ends up getting rid of the loop’s scope too soon. It’d be best to fix that I guess.

superaxander:
Fixed in #1408

PR #1379, opened 2025-07-30
Quality of life improvements on extract

sakehl:
Checklist: 

# PR description
Some quality of life changes for 'extract'

Now with for-loops, you get something more useful. For for loop `for(int i = A; C; ...)` I've added that `i == A` as precondition to the extracted method. And I've added `!C` to the post condition of the extracted method.

For a `while(C)`, I've added `!C` to the post condition of the extracted method.

Also I've added a way to add 'extract decreases` as a way to mark the extracted method as terminating.


sakehl:
Mostly see https://github.com/utwente-fmt/vercors/pull/1379/files#diff-05f1dfebde28bb768466c1ca7cc0c896725e54d725b82425ff799d661cd967ee

For examples.

PR #1382, opened 2025-09-09
Remove in-post compare for mixing PVL and LLVM

RobertMensing:
Removes the in-post comparison that was used for mixing PVL contracts and LLVM. 
The allowed type equalities are now explicitly listed instead of using the dispatch for this. Using dispatch caused issues because types got declared while no scope was available.


superaxander (approved):
LGTM

PR #1397, opened 2026-02-04
Gpu kernel improv 1

sakehl:

# PR description

* Adds the option for GPU bodies to be automatically extracted (makes it quicker to verify most gpu kernels) via `extract_body` annotation.
* Measure backend time via option `--dev-time-backend`. Usefull for reporting the verification time of viper.
* Include bool definition in opencl, so you do not have to include bool.h (which is how opencl works)
* Changed flag in C preprocessor, such that `#define UNIQUE /*@unique<1>*/` works (comments are added with inlined define things)
* Some improvements for the simplify quantifier rewrite. 
  * Gets more info from barrier contracts
  * Gets info from A to use in B in situation:  `A ==> B`
  * Nest foralls better when an implies is there, with par block encoder.
* Allow let definitions inside triggers, by inlining such lets.
* If you scale like this `[nb][nt]Perm(xs[i], r)`, it turns into `Perm(xs[i], Scale(nb)*Scale(nt)*r)` where Scale is a pure function, with a precondition added that the input should be >0, leads to incompleteness somehow. However, when you write `nb*nt*r`, this does not happen (I do not know why).
  * So I added support, that if the contract states nt>0 or something, we can simply write nt instead of Scale(nt), since we do not have to check the precondition.
* Some better parsing for `extract` and `frame` in C with annotations:
  * `/*@ extract frame */ /@* ensures A;*/ /@*{*/ B /@*}*/`  is now possible, which was not.  
* There were some type errors related to type modifiers. So I say we strip type annotations, for type checking numeric bin operators.
* Allow for int[5][2] to be declared as shared memory inside GPU kernels, was not possible before.
* Allow requires \old(xs[i]) in GPU kernel contracts. This is translated to 3 contracsts: Function contract, outer parallel block, inner parallel block. For function contract the old is removed, for outer and inner the old stays there. This is useful when you do not use `--no-infer-heap-context-into-frame` to say something like `xs[i]==\old(xs[i])`
  * Also allows for `(\forall i=0..n; \old(xs[i]) == seq_xs(i))`, to record the old value of xs in a sequence, which is otherwise not possible.
* Simplify \ltid and \gtid better when we know a GPU kernel only has 1 or 2 dimensions. This leads to more predictable behaviour, especially in barriers when using these constructs.

I think this covers all changes. These were mostly all needed for the latest paper submission, Marieke, Anton and me did.


src/col/vct/col/ast/expr/ambiguous/AmbiguousPlusImpl.scala
sakehl, line 1: This looks strange, but if you look at the complete file this was present twice.

src/col/vct/col/ast/expr/misc/OldImpl.scala
sakehl, line 1: allow old in gpu kernels

src/col/vct/col/ast/expr/op/BinExprImpl.scala
sakehl, line 1: strip type modifiers for numeric bin operators, to determine type.

src/col/vct/col/ast/lang/c/CFunctionDefinitionImpl.scala
sakehl, line 1: Here we notify the checker that we are (or not) in a gpu kernel

src/col/vct/col/ast/util/ExpressionEqualityCheck.scala
```diff
@@ -93,158 +95,186 @@ object ExpressionEqualityCheck {
...
-          val bmgr = fmgr.getBooleanFormulaManager
-          val imgr = fmgr.getIntegerFormulaManager
-
-          def tdiv(a: Expr[G], b: Expr[G]): Expr[G] =
-            Select(
-              a >= const(0) || a % b === const(0),
-              a / b,
-              a / b + Select(b > const(0), const(1), const(-1)),
-            )
+        Using(ctx.newProverEnvironment(ProverOptions.GENERATE_MODELS)) {
```
sakehl, line 98: generate models allows for easier debugging, since you get back the model of the SAT solver

src/col/vct/col/ast/util/ExpressionEqualityCheck.scala
```diff
@@ -93,158 +95,186 @@ object ExpressionEqualityCheck {
...
-                for {
-                  b1 <- addInt(e1); b2 <- addInt(e2)
-                } yield imgr.greaterOrEquals(b1, b2)
-              case BooleanValue(b) => Some(bmgr.makeBoolean(b))
-              case t: SymbolicTerm[G] if isBool(e) && isValidSymbolicTerm(t) =>
-                if (varBoolMap.contains(t))
-                  Some(varBoolMap(t))
-                else {
-                  val x = bmgr.makeVariable(s"b$id")
+            def getName(t: SymbolicTerm[G]): String = {
```
sakehl, line 131: get nicer name, for debugging SMT queries

src/col/vct/col/ast/util/ExpressionEqualityCheck.scala
```diff
@@ -537,17 +567,50 @@ class ExpressionEqualityCheck[G](info: Option[AnnotationVariableInfo[G]]) {
...
+          isConstantIntRecurse(right).map(l => return Some(l))
+        case Greater(left, right) if !isLower && right == e =>
+          isConstantIntRecurse(left).map(l => return Some(l + 1))
+        case GreaterEq(left, right) if !isLower && right == e =>
+          isConstantIntRecurse(left).map(l => return Some(l))
+
+        case Eq(left, right) if left == e =>
+          isConstantIntRecurse(right).map(l => return Some(l))
+        case Eq(left, right) if right == e =>
+          isConstantIntRecurse(left).map(l => return Some(l))
```
sakehl, line 593: this section adds a better bound for getBound(e, isLower).
For instance if the usefull conditions say left == e, and left can be written as an integer l, then a good bound for e is l. (both an upper as a lower bound).

src/parsers/vct/parsers/parser/ColCParser.scala
sakehl, line 1: This options allow comments to be added with `#define` things in C.

src/rewrite/vct/rewrite/lang/LangCToCol.scala
```diff
@@ -1175,6 +1202,54 @@ case class LangCToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+                          val post = Star(
+                            context,
+                            sub.dispatch(rw.dispatch(contractEnsures)),
+                          )
+                          val inner = rw.dispatch(implFiltered.get)
+                          Extract(
+                            FramedProof(pre, inner, post)(PanicBlame("TODO")),
+                            None,
+                          )(PanicBlame("TODO: Decreases"))
+                        } else { rw.dispatch(implFiltered.get) }
```
sakehl, line 1272: This part adds a framed block for the inner GPU thread. Should probably fix TODO here.

src/rewrite/vct/rewrite/lang/LangCToCol.scala
```diff
@@ -1914,12 +2007,37 @@ case class LangCToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+    var result: Set[Expr[G]] = Set()
+
+    for (l <- pres) {
+      if (!isContext(l))
+        result += l
+    }
+    for (r <- remainingPosts)
+      result += r
+    result
+  }
```
sakehl, line 2064: This was a fix needed for context annotations in barriers. Since for context annotations we do not need a memory fence. But is was erroring when there was a forall.

src/rewrite/vct/rewrite/lang/LangCToCol.scala
```diff
@@ -2446,22 +2586,20 @@ case class LangCToCol[Pre <: Generation](rw: LangSpecificToCol[Pre])
...
+    if (!isLocalThreadIdConstant(2, o)) {
+      if (!isLocalThreadIdConstant(1, o)) {
+        base =
+          base +
+            getCudaLocalSize(0, o) *
+            getCudaLocalSize(1, o) * getCudaLocalThread(2, o)
+      } else { base = base + getCudaLocalSize(0, o) * getCudaLocalThread(2, o) }
+    }
+    // get_local_id(0) + get_local_id(1)*get_local_size(0) + get_local_id(2)*get_local_size(1)*get_local_size(0)
+    base
```
sakehl, line 2631: simplify local thread id, when things are constant.

src/rewrite/vct/rewrite/lang/LangSpecificToCol.scala
```diff
@@ -574,7 +574,7 @@ case class LangSpecificToCol[Pre <: Generation](
 
       case cmp: AmbiguousComparison[Pre] => c.rewriteComparison(cmp)
       case ord: AmbiguousOrderOp[Pre] => c.rewriteComparison(ord)
-
+      case old: Old[Pre] if c.inRequiresOfKernel => dispatch(old.expr)
```
sakehl, line 577: if in requires of GPU kernel, remove the old.

src/rewrite/vct/rewrite/ExtractInlineQuantifierPatterns.scala
```diff
@@ -93,10 +93,13 @@ case class ExtractInlineQuantifierPatterns[Pre <: Generation]()
...
-            Pattern(i.group, i.inner, letBindings.top.toSeq.toMap)
-        }
-        dispatch(i.inner)
+          val pattern = Pattern(i.group, i.inner, letBindings.top.toSeq.toMap)
+          patterns.toSeq(i.parent) += pattern
+          // Dispatch for the inner, maybe it contains more triggers?
+          dispatch(i.inner)
+          // But return the made pattern, which inlined the lets.
+          pattern.make()
+        } else { dispatch(i.inner) }
```
sakehl, line 102: inline lets in patterns

src/rewrite/vct/rewrite/ParBlockEncoder.scala
```diff
@@ -156,26 +156,35 @@ case class ParBlockEncoder[Pre <: Generation]() extends Rewriter[Pre] {
...
+                    range ==> scale(dispatch(body)),
+                  )(s.blame)(body.o)
+                case other =>
+                  Starall(
+                    variables.dispatch(quantVars),
+                    Nil,
+                    range ==> scale(dispatch(other)),
+                  )(ParBlockNotInjective(block, other))(other.o)
+              }
+            implies.map(i => Implies(dispatch(i._1), res)(i._2)).getOrElse(res)
```
sakehl, line 187: this nests forall better, when an implies is there with a parblock

src/rewrite/vct/rewrite/ResolveScale.scala
sakehl, line 1: This makes us not use a scale function, when not needed (e.g. for scale(n) is not needed if know that n>0).

src/rewrite/vct/rewrite/SimplifyNestedQuantifiers.scala
```diff
@@ -86,6 +87,23 @@ case object SimplifyNestedQuantifiers extends RewriterBuilder {
           " because the quantifier variables are not partitioned."
       )
   }
+
+  private class MyCtx extends Ctx {
```
sakehl, line 91: Nicer names for forall info about triggers missing or other errors

src/rewrite/vct/rewrite/SimplifyNestedQuantifiers.scala
```diff
@@ -230,18 +256,28 @@ case class SimplifyNestedQuantifiers[Pre <: Generation]()
...
+        val left = dispatch(e.left)
         topLevel = false
-        equalityChecker = ExpressionEqualityCheck(
-          Some(infoGetter.finalInfo(constantInfo.toSeq.flatten))
-        )
-        rewriteBinder(e)
+        val right = dispatch(e.right)
+        // Reset the scoped info from the implies
+        infoGetter = oldInfoGetter
+        Implies(left, right)(e.o)
```
sakehl, line 280: Get info from implies

src/rewrite/vct/rewrite/SimplifyNestedQuantifiers.scala
```diff
@@ -313,8 +356,10 @@ case class SimplifyNestedQuantifiers[Pre <: Generation]()
 
   override def dispatch(stat: Statement[Pre]): Statement[Post] = {
     stat match {
-      case Exhale(e) =>
-      case Inhale(e) =>
+      case Exhale(_) =>
+      case Inhale(_) =>
+      case Assert(_) =>
+      case Assume(_) =>
```
sakehl, line 362: also add info for asserts and assume statements, so rewrite is better.

sakehl:
I have added some comments, so a review is little bit easier (its a lot I know!)

superaxander:
Next time you post a bunch of comments I recommend starting a review, adding the comments, and then posting the review all at once. Right now I got an email for every comment you left :P

I will review this tomorrow :)

sakehl:
Thanks a lot! And apologies, didn't know that.

src/col/vct/col/ast/expr/misc/OldImpl.scala
superaxander: What does old mean in the precondition of a kernel?

sakehl:
I've added two examples (xtrsv.cl and xswap.cl) which heavily rely on these changes.

However xtrsv seems to be quite slow, which it was not before (previously the backend took about 25s, now it is more like 120s). I suspect this is because the newer Viper version, but unsure.

src/col/vct/col/ast/expr/misc/OldImpl.scala
superaxander: Ok I just saw that later on you remove the olds in the precondition of a kernel. Why allow them if they would just get removed anyway?

src/rewrite/vct/rewrite/SimplifyNestedQuantifiers.scala
```diff
@@ -39,7 +40,7 @@ import scala.annotation.nowarn
   * j; 0 <= j < 5; xs[j])
   *
   * and if a quantifier isn't in the "body" of the forall. E.g. forall(i,j: Int.
-  * 1 <= i && i< n && 0 < j; xs[j]>0) ====> n > 1 ==> forall(j: Int; 0 < j;
+  * 1 <= i && i< n && 0 < j; xs[j]>0) ====> n > 1 ==> Fforall(j: Int; 0 < j;
```
superaxander: Typo?

superaxander (dismissed):
Looks good to me

src/col/vct/col/ast/expr/misc/OldImpl.scala
sakehl: See `examples/concepts/gpgpu/xtrsv.cl` 

Here i write
`context \ltid < n ==> \old(x[acc1d(\ltid, x_offset, n, x_inc)]) == {:_x[\ltid]:};`
Even though x is written to by the threads. It has permission:
`context \ltid < n ==> Perm({:x[acc1d(\ltid, x_offset, n, x_inc)]:}, write);`

So the old is used in parallel block contract of the thread block, and in the parallel block contract of the the threads. It is only removed in the function for the whole kernel.

sakehl (commented):
I changed the todo blames, made the parsing of extract_body a little bit better, added some tests for extract_body. So I think we are there!

src/col/vct/col/ast/expr/misc/OldImpl.scala
sakehl: Maybe better, you can even writen (forall int i; 0<=i<n; \old(xs[i]) == ...) so you record the values of the complete array, even though it might be altered later on.

superaxander: Even for the parallel block contract it doesn't make a lot of sense to me what the \old means in its precondition.

sakehl:
I still wanted to give you an example when old is usefull as precondition. This following example shows this.
Here, I use the option `--no-infer-heap-context-into-frame` (which sometimes you might want to use). And you need to old annotation to successfully verify

```opencl
#include "opencl.h"

/*@
  context get_work_dim() == 1;
  context get_local_size(0) == 2;
  context get_num_groups(0) == 1;

  context in != NULL && out != NULL;
  context \pointer_length(in) == 2;
  context \pointer_length(out) == 2;
  context Perm({:in[\gtid]:}, 1\2);
  context Perm({:out[\gtid]:}, write);
  // Because of the following annotations using old, this verifies using --no-infer-heap-context-into-frame
  requires {:in[\gtid]:} == \old(in[\gtid]);
  ensures {:out[\gtid]:} == \old(in[\gtid]);
@*/
__kernel void blur_x(global int* in, global int* out) {
  int tid = get_group_id(0) * get_local_size(0) + get_local_id(0);
  out[tid] = in[tid];
}
```

superaxander:
Ah right I see, it's because of the encoding as a loop which causes the heap to be havoced. We should really fix that bug that requires `--no-infer-heap-context-into-frame` at some point but it's a tricky one.

Issue #1403, opened 2026-02-17
\unfolding disallowed in pure postcondition


ArmborstL:
Resources are disallowed in postconditions of pure functions, so `ResourceTermImpl` raises a `ResourceInPostcondition` error.
However, this is too restrictive: expressions of the form `\unfolding res() \in expr` should be allowed, but also raise the `ResourceInPostcondition` error.
Example:
```java
class T {
  int x;

  //@ resource s() = Perm(this.x, 1\2);

  //@ requires t.s();
  //@ ensures \result == (\unfolding t.s() \in t.x);
  //@ pure
  int test(T t) {
    //@ unfold t.s();
    return t.x;
  }
}
```
gives `Resource terms may not appear in the postcondition of functions (a.k.a pure procedures)` for the `t.s()` in the `\unfolding`

PR #1404, opened 2026-02-17
workaround issue #1403

ArmborstL:
- [X] No changes to wiki required.

# PR description
Quick-fix for issue #1403, filtering the `ResourceInPostcondition` error if it's coming from an `\unfolding` target.
Not solving the issue for e.g. scaled expressions like `\unfolding [1\2]res() \in expr`.


ArmborstL:
@superaxander The failing test in ScalaTest MATRIX[5] (relating to LLVM Spec in `struct.c`) seems unrelated to my changes, right? So I guess we can merge this PR then?

superaxander:
Yes although I think it would be safe to simply ignore the ResourceInPostcondition error only for the `res` part of the unfolding. I.e. instead of calling checkContextRecursor you call check on the two children individually and filter out the error from the `res.check`

ArmborstL:
@superaxander I think this is what you suggested in our discussion, right? 
It now also works for `\unfolding [scale]res() \in expr`.

superaxander:
Indeed, I'll merge it 👍

PR #1420, opened 2026-06-19
Fix var to pointer

RobertMensing:
Make the VariableToPointer-pass generate LocalHeapVariables for the generated intermediaries.
Generating plain variables causes issues with missing permissions when ByValueClasses are passed as arguments to functions.

### DIFF
--- a/src/rewrite/vct/rewrite/VariableToPointer.scala
+++ b/src/rewrite/vct/rewrite/VariableToPointer.scala
@@ -35,6 +35,11 @@ case object VariableToPointer extends RewriterBuilder {
         "Taking the address of a pure function's argument is not supported"
       )
   }
+
+  object NonNullAlwaysInitialized
+      extends PanicBlame(
+        "Non-null pointer should always be initialized successfully"
+      )
 }
 
 case class VariableToPointer[Pre <: Generation]() extends Rewriter[Pre] {
@@ -47,7 +52,7 @@ case class VariableToPointer[Pre <: Generation]() extends Rewriter[Pre] {
 
   val addressedSet: mutable.Map[Node[Pre], PointerSort] =
     new mutable.HashMap[Node[Pre], PointerSort]()
-  val variableMap: SuccessionMap[Variable[Pre], Variable[Post]] =
+  val variableMap: SuccessionMap[Variable[Pre], LocalHeapVariable[Post]] =
     SuccessionMap()
   val noTransform: ScopedStack[scala.collection.Set[Variable[Pre]]] =
     ScopedStack()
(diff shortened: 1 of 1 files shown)


src/rewrite/vct/rewrite/VariableToPointer.scala
```diff
@@ -142,32 +150,36 @@ case class VariableToPointer[Pre <: Generation]() extends Rewriter[Pre] {
...
-                    Some(Scope(locals, block)(proc.o))
+                  // Add declarations & assignments for the localHeapVariables to the start of the body
+                  localHeapVariables.scope {
+                    variables.scope {
+                      val locals =
+                        localHeapVariables.collect {
+                          extraVars.map { case (_, pointer, _) =>
+                            localHeapVariables.declare(pointer)
+                          }
+                        }._1
```
superaxander: I think declaring the local heap variables separately is not needed here. In general it is always use either `variables.declare`/`localHeapVariables.declare` *or* use `LocalDecl`/`HeapLocalDecl` never both. Since you are calling `localHeapVariables.collect` here this is a roundabout way of writing `extraVars.map { case (_, pointer, _) => pointer }`

src/rewrite/vct/rewrite/VariableToPointer.scala
```diff
@@ -142,32 +150,36 @@ case class VariableToPointer[Pre <: Generation]() extends Rewriter[Pre] {
...
+                      val localHeapDecls = locals.map { case v =>
+                        HeapLocalDecl(v)(v.o)
+                      }
+                      val block =
+                        Block(localHeapDecls ++ extraVars.map {
+                          case (normal, pointer, Normal(_)) =>
+                            Assign(
+                              pointer.get(NonNullAlwaysInitialized)(normal.o),
+                              normal.get(normal.o),
+                            )(AssignLocalOk)(proc.o)
```
superaxander, line 166: There is a helper for this in ASTBuildHelpers, so you can write `assignLocal(pointer.get(NonNullAlwaysInitialized)(normal.o), normal.get(normal.o))(proc.o)`

src/rewrite/vct/rewrite/VariableToPointer.scala
RobertMensing: Thanks! I simplified it.

src/rewrite/vct/rewrite/VariableToPointer.scala
RobertMensing: Thanks for the hint! assignLocal unfortunately did not work out because the get adds a DerefPtr around the Local.

PR #1429, opened 2026-08-18
Integrating an LSP into VerCors

superaxander:
Checklist: 

# PR description
This PR integrates an LSP server into VerCors for use with the IDE plugin (see https://github.com/utwente-fmt/vercors-ide/pull/59). 
This is mostly Yusof's work but I've cleaned it up such that log messages are not commented out and (except for UserErrors) no reflection is needed to get the origin for the diagnostic messages.

Future work for the plugin:
- Multi-file support (currently everything assumes there is only one file which causes weird behaviour when another file is involved, for example the files that are imported from the res/universal/res/adt folder or included with `#include` statements in C)
- While the interface currently supports adding multiple VerCors installations (for working with different versions) only the most recently added one is used. We should decide if we want to keep this functionality and if so implement it correctly.
- Currently we need a window reload after adding the VerCors path, that could be improved
- We should add a way to specify which java binary to use and which JVM flags (I had to add `--enable-native-access=ALL-UNNAMED` to get it to run on JDK 19 and up, but it that flag is not known for anything lower than that. If we switch to an executable jar we can move that flag to the manifest and avoid needing different JVM flags for different Java versions)
- Currently the progress highlighting isn't cleared when the file is done being verified
- Legibility of the errors can be improved, currently they seem to be abbreviated too much
- The settings for the options is a bit broken, especially the custom flags which are broken when they're empty (it sends an empty argument to VerCors which is not a valid CLI option)
- Ideally we query what settings are supported through the LSP
- Currently we still use reflection to look for origins in UserErrors, that should be replaced with some function on the UserError trait.


build.sc
```diff
@@ -570,7 +570,14 @@ object vercors extends Module {
 
     def deps = Agg(
       ivy"com.github.scopt::scopt:4.0.1",
+      ivy"org.eclipse.lsp4j:org.eclipse.lsp4j:1.0.0",
+      ivy"io.circe::circe-core:0.14.5",
```
bobismijnnaam: Don't we already have a JSON parser? I remember using it for the JavaBIP support. Though maybe that was too specialized and we need two json libraries?

build.sc
```diff
@@ -570,7 +570,14 @@ object vercors extends Module {
...
+      ivy"io.circe::circe-generic:0.14.5",
+      ivy"io.circe::circe-parser:0.14.5",
     )
+    override def scalacOptions = T {
+      Seq("-Xmixin-force-forwarders:false")
```
bobismijnnaam: Why is this necessary? Might be helpful to put a comment there in case it's crucial for some reason

src/main/vct/main/stages/Resolution.scala
```diff
@@ -41,8 +41,12 @@ import java.io.{
 import java.nio.charset.StandardCharsets
 import java.nio.file.Path
 
+trait HasCheckErrors {
```
bobismijnnaam: Is this a necessary abstraction? Can VerificationErrorsUtils not pattern match on Resolution.InputResolutionError? Or do you expect many implements of HasCheckErrors? There seems to be only one atm

src/main/vct/lsp/MyLanguageServer.scala
```diff
@@ -0,0 +1,64 @@
...
+  TextDocumentService,
+  WorkspaceService,
+}
+
+import java.util.concurrent.CompletableFuture
+
+class MyLanguageServer extends LanguageServer {
+  private val textDocumentService = new MyTextDocumentService()
+  private val workspaceService = new MyWorkspaceService()
+  MyLanguageServer.client = null
```
bobismijnnaam, line 17: I guess the singleton pattern is mandated by the library and can't be avoided?

res/universal/res/lsp/language-server-java-c-matches.json
bobismijnnaam, line 1: Is this (and language-server-pvl-matches.json) auto generated, or manually maintained? If the second we should document this link somewhere, e.g. in the respective grammers maybe? Or is it already documented somewhere else? If it's automatically generated, can it be made part of the regular vercors build step? If neither is possible, can we add a check somewhere (ci step, git hook, build check) to make sure this list and the grammar don't diverge? Or do you have no more time and is this simply a limitation we have to accept (also a fair answer :))

mill-build/util/src/util/ScalaModule.scala
bobismijnnaam, line 1: Maybe drop this edit, and the VercorsModule.scala and mill-build/build.sc edits too?

build.sc
bobismijnnaam: (I also noticed gson imported in one of the files)

res/universal/res/lsp/language-server-java-c-matches.json
superaxander: Yusof's thesis doesn't mention it so I assume it was a manual thing

src/main/vct/lsp/MyLanguageServer.scala
superaxander: I think so yeah

src/main/vct/main/stages/Resolution.scala
superaxander: When I was looking at the code for this I sort of assumed that he would've had multiple things for it to implement but yeah looking at it now only one thing implements it.

build.sc
superaxander: Yeah this actually caused me quite a headache to figure out again, I put the explanation in the commit, but I'm also still thinking of ways to work around this. It's due to the LSP4J's use of reflection in combination with Scala's inaccurate marking of methods as synthetic.

build.sc
superaxander: Yeah it seems we already have upickle for this. The gson imports seem to be there because LSP4J uses GSON

mill-build/util/src/util/ScalaModule.scala
superaxander: I don't think that's possible. Since there was no newline at the end of these files and the files were modified and then reverted it will always show as a change (unless I squash the commit history). If I add a new line then that will not be an issue again in the future.

superaxander: What it looks like if I don't add the new line:
Diff showing that an identical line was removed and added

res/universal/res/lsp/language-server-java-c-matches.json
bobismijnnaam: Oh no, well, too bad

